PR #20, opened 2012-09-24
improved install langs

Kamahl19:


root/socialnet/install_sn.php
```diff
@@ -51,7 +51,6 @@
  */
 
 $user->add_lang(array('ucp', 'mods/socialnet'));
-$language_file = 'mods/socialnet_acp';
```
senky, line 54: why did you remove language file?

Kamahl19, line 54: because it is not used anywhere

senky, line 54: Alright

senky:
MA

root/socialnet/install_sn.php
Culprit-CZ, line 54: Disagree, language file is used for lang to ADMIN EVENT LOG

Culprit-CZ:
Kde je description?????

Jinak bez souboru to bude v ACP -> System -> Admin log davat spatne infromace

root/socialnet/install_sn.php
Kamahl19, line 54: ah, you are right

root/socialnet/install_sn.php
```diff
@@ -30,7 +30,7 @@
 }
 
 // The name of the mod to be displayed during installation.
-$mod_name = 'Social Network';
+$mod_name = 'phpBB Social Network';
```
senky, line 33: I do not like it does not use language file. Or it is not able to be used here?

root/socialnet/update_sn_0.7.0.php
```diff
@@ -30,7 +30,7 @@
 }
 
 // The name of the mod to be displayed during installation.
-$mod_name = 'Social Network - update';
+$mod_name = 'phpBB Social Network update';
```
senky, line 33: the same - I do not like it does not use language file (if it can)

Kamahl19, line 33: Yes we can use lang files in there, but we have never used them for umil files. However you are right, we should use it.

root/socialnet/install_sn.php
Culprit-CZ, line 54: Better to use instead 2 lines 53 and 54 ony 1

``` php
$language_file = array('ucp', 'mods/socialnet', 'mods/socialnet_acp');
```

root/socialnet/update_sn_0.7.0.php
senky, line 33: So do not forget about assigning this variable **after** language file is loaded ;)

Culprit-CZ:
Tys to jeste neudelal?

senky:
Replaced with #31

PR #5, opened 2012-09-25
[t/4068] Opening one of tabs in profile in a new window will result in not properly styled page.

senky:
When clicking on one of 4 tabs in profile (Activity, Info, Friends, Statistics) in the way it should open in a new window (ie. by clicking when holding Ctrl, or by clicking middle button on a mouse), tab will open in new window, however not properly styled.

This happens because modes "wall", "infor", "friends" and "stats" of profile.php (?mode=xx) do not return full page, but only that part, that is required to be displayed in a tab.

To fix this:
1. simple way:
   profile.php with specified mode should check if request comes from ajax call, or common browser call ($_SERVER['HTTP_X_REQUESTED_WITH'] header). This header is, however, not present in all servers.
2. more safe way:
   send new parameter by ajax, identifying itself. Otherwise, send full page to the user.

http://phpbbsocialnetwork.com/viewtopic.php?f=60&t=4068

### DIFF
--- a/root/socialnet/js/m.profile.js
+++ b/root/socialnet/js/m.profile.js
@@ -1,10 +1,10 @@
 /**
- * 
+ *
  * @package phpBB Social Network
  * @version 0.7.0
  * @copyright (c) 2010-2012 Kamahl & Culprit http://phpbbsocialnetwork.com
  * @license http://opensource.org/licenses/gpl-license.php GNU Public License
- * 
+ *
  */
 (function($) {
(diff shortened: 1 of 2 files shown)


Culprit-CZ:
A ted co? Ja nevim, protoze jsem to delal ja.
Ja bych to Mergnul

senky:
Ak som na nic nezabudil, tak napis do komentaru:

MA

a pockajme na Kamahla.

PR #23, opened 2012-09-26
SN expander read more/read less margins

Culprit-CZ:
http://phpbbsocialnetwork.com/viewtopic.php?f=59&t=4097

``` css
.ltr .sn-expander-more a,.ltr .sn-expander-less a {
    float: right;
    margin-right: 5px;
}

.rtl .sn-expander-more a,.rtl .sn-expander-less a {
    float: left;
    margin-left: 5px;
}
```


senky:
MA

PR #7, opened 2012-09-26
Add note about limits of curl (used in fetching page for status)

senky:
Having safe_mode enabled or open_basedir set, curl results in this error:

```
Warning: curl_setopt_array() [function.curl-setopt-array]: CURLOPT_FOLLOWLOCATION cannot be activated when safe_mode is enabled or an open_basedir is set
```

Therefore it would be fine to include some kind of note about it in author's note or DIY instructions, for users to know that limit.

### DIFF
--- a/install_instructions/install_mod.xml
+++ b/install_instructions/install_mod.xml
@@ -16,6 +16,7 @@ Changelog: http://phpbbsocialnetwork.com/viewtopic.php?t=1023
 phpBB Social Network Facebook page: http://www.facebook.com/pages/phpBB-Social-Network/180271885389370
 
 Feel free to ask for support or request more features here http://phpbbsocialnetwork.com/]]></description>
+		<author-notes lang="en"><![CDATA[Please note, that if you want to use "Fetch page" feature (while posting new status), you need to have CURL enabled, safe_mode off and open_basedir not set.]]></author-notes>
 		<author-group>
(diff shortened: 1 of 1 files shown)


install_instructions/install_mod.xml
```diff
@@ -15,56 +15,57 @@ Changelog: http://phpbbsocialnetwork.com/viewtopic.php?t=1023
...
-		<open src="search.php">
-			<edit>
-				<find><![CDATA[$zebra[($row['friend']) ? 'friend' : 'foe'][] = $row['zebra_id'];]]></find>
-				<action type="replace-with"><![CDATA[$zebra[($row['foe']) ? 'foe' : 'friend'][] = $row['zebra_id'];]]></action>
-			</edit>
-		</open>
-		<open src="viewonline.php">
-			<edit>
+Feel free to ask for support or request more features here http://phpbbsocialnetwork.com/]]></description>
+		<author-notes lang="en"><![CDATA[Please note, that if you want to use "Fetch page" feature (while posting new status), you need to have CURL enabled, safe_mode off and open_basedir not set.]]></author-notes>
```
senky, line 19: Only this line was changed. I do not know why git diffed also other lines. I uploaded install_mod.xml to online MOD creator, added author notes and downloaded it. Nothing more.

PR #35, opened 2012-10-01
SN theme compiler

Culprit-CZ:
Compiler allows more effective create any CSS file for phpBB Social Network.

Allow to use many things that are allowed using lessphp.

Otherwise we use now this compiler. But is included into phpBB folder.
Here is out of root folder of phpBB.


sn-theme/hooks/hook_sn-theme.php
```diff
@@ -0,0 +1,132 @@
...
+		$_out = $css->print->plain();
+		$_out = preg_replace('/url\(([^)]+)\)/si', 'url("\1")', $_out);
+		$_out = preg_replace("/:([^;\n]+);/si", ': \1;', $_out);
+		$_out = preg_replace("/\n{\n/si", " {\n", $_out);
+		file_put_contents($less_output_file, $_out);
+	}
+	catch (exception $ex)
+	{
+		$message = file_get_contents($compiler_path . 'error_body.html');
+		$message = preg_replace('/\{ERROR_MESSAGE\}/s', $ex->getMessage(), $message);
```
senky, line 128: I think we can use simple `str_replace()` here.

PR #38, opened 2012-10-03
SN Theme Generator

Culprit-CZ:
#### Theme Compiler

Compiler allows more effective create any CSS file for phpBB Social Network.
Allow to use many things that are allowed using [lessphp](http://leafo.net/lessphp/docs/).

Otherwise we use now this compiler. But is included into phpBB folder.
Here is out of root folder of phpBB.

---
#### Why again?

See #35
Kamahl has merged and closed.
But in branch [develop](https://github.com/phpBB-Social-Network/phpBB-Social-Network/tree/develop) is nothing.


sn-theme/hooks/hook_sn-theme.php
```diff
@@ -0,0 +1,132 @@
...
+		$css->set_cfg('allow_html_in_templates', FALSE);
+		$css->set_cfg('template', 'low');
+		$css->set_cfg('compress_colors', TRUE);
+		$css->set_cfg('sort_properties', TRUE);
+		$css->set_cfg('sort_selectors', false);
+		$css->set_cfg('compress_font-weight', false);
+
+		$css->parse($compile);
+		$_out = $css->print->plain();
+		$_out = preg_replace('/url\(([^)]+)\)/si', 'url("\1")', $_out);
```
senky, line 124: Plase, give those 3 lines inline comments on what this preg_replace does.

Culprit-CZ, line 124: 108-116<b>::</b> Set the [CSSTidy](http://csstidy.sourceforge.net/usage.php) to be CSS optimized such as phpBB
118<b>::</b> Optimize the Cascade Style Sheet
119<b>::</b> Take the plain CSS not the HTML output
120<b>::</b> Because in some way is removed `"` from URLs with `preg_replace` I get the `"` back

senky, line 124: Yes, write this to code itself, so that anyone in any time can understand the code, please.

Kamahl19:
What is this for?

PR #40, opened 2012-10-03
'IM_ALLOW_SOUND' and 'IM_ALLOW_SOUND_EXPLAIN' should be moved to socialnet.php

senky:
```
'IM_ALLOW_SOUND' => 'Play sound when receive messages',
'IM_ALLOW_SOUND_EXPLAIN' => 'This option enables/disables sound when receiving a new message',
```

This should be in /socialnet.php instead of socialnet_acp.php (becouse it isnt acp)

http://phpbbsocialnetwork.com/viewtopic.php?f=62&t=4153


Culprit-CZ:
MA

PR #42, opened 2012-10-04
Add documentation to root/socialnet/includes/functions.php

senky:
Adding phpDocumentor2 comments before each function, class and method in root/socialnet/includes/functions.php


Culprit-CZ:
MA

PR #41, opened 2012-10-05
Improve issue2pr.php so that it can guess issue ID

senky:
To effectivelly develop on github, we create issue for each code change. To issue, we attach code and then it becomes pull request. We should create issue first, and after that create branch with its name (`i/<issue id>`) and send pull request from this branch for other developers to easy check the code, etc.

Thus, branches include issue id that can be easilly guessed by script.

I propose changing issue2pr.php, so that if it can load branch name, and branch name starts with `i/` followed by numbers only, it should type this number into field called "Issue ID".


Culprit-CZ:
MA

PR #46, opened 2012-10-08
Adding new menu item in ACP which title starts with "/" results in redirecting to index of board

senky:
in ACP -> Social Network -> Blocks configuration -> Menu block try to input `/test` in form. Submit and you will be redirected to index.php of your board.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=2674


Culprit-CZ:
MA

PR #52, opened 2012-10-08
Add documentation to root/socialnet/includes/sn_core_addons.php

senky:
Adding phpDocumentor2 comments before each function, class and method in root/socialnet/sn_core_addons.php

### DIFF
--- a/root/socialnet/includes/sn_core_addons.php
+++ b/root/socialnet/includes/sn_core_addons.php
@@ -1,25 +1,53 @@
 <?php
 /**
+ * SN Core Addons
  *
- * @package phpBB Social Network
- * @version 0.7.0
- * @copyright (c) phpBB Social Network Team 2010-2012 http://phpbbsocialnetwork.com
- * @license http://opensource.org/licenses/gpl-license.php GNU Public License
+ * @todo add "since", verify authors
+ *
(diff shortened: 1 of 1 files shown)


root/socialnet/includes/sn_core_addons.php
```diff
@@ -1,25 +1,53 @@
...
+ */
 if (!defined('SOCIALNET_INSTALLED') || !defined('IN_PHPBB'))
 {
 	return;
 }
 
+/**
+ * Core Addons class
+ *
+ * @package phpBB-Social-Network
```
Culprit-CZ, line 27: Creating another package. One is defined on line 10
I think that should be the same names

senky, line 27: Ok, can we use `phpBB-Social-Network`? I think using dashes (`-`) gives us much more advantages (in exporting data by way that does not allow spaces, etc.).

Culprit-CZ, line 27: Without dashes looks more better

Kamahl19:
Is this already finished?

Culprit-CZ:
No is not finished, but as first documentation for this file.
MA

PR #48, opened 2012-10-09
Confirm box shadow does not move if draggable enabled

Culprit-CZ:
http://phpbbsocialnetwork.com/viewtopic.php?f=62&t=4184

### DIFF
--- a/root/socialnet/js/m.confirmbox.js
+++ b/root/socialnet/js/m.confirmbox.js
@@ -1,7 +1,7 @@
 /**
 *
 * @package phpBB Social Network
-* @version 0.6.3
+* @version 0.7.0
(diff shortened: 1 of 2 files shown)


senky:
What lines was changed really? This PR tries to change a really big block of code, what, at least for me, is unacceptable.

Culprit-CZ:
`m.confirmBox.js` is correctly displayed. There was some smaller changes by declaration.
`m.socialnet.js` was mostly changed at this part

```
confirmBox: {
}
```

Than I move function `dropShadow` to `confirmBox:{}`

That is the main reason why the file is mostly marked as changed

senky:
Alright, if 668dbcd belongs to this PR too: MA

Culprit-CZ:
Should belongs to this PR

PR #57, opened 2012-10-09
Profile redirect variable

Culprit-CZ:
Use a global variable is not an advantage

> ``` php
> $_SERVER['PHP_SELF']
> ```

Use instead 

> ``` php
> "{$phpbb_root_path}profile.$phpEx"
> ```


senky:
Except for changes in `root/profile.php`, there should be no file included in this PR.

Culprit-CZ:
I don't know why is the first commit assigned to this PR.
That appear for me also in another PR.

senky:
Please, remove it, we cannot merge same changes more times. It will result at best in merge conflict.

PR #50, opened 2012-10-09
Comment box does not collapse after blur

Culprit-CZ:
When you focus on comment box, it will enlarge. However, after you blur it, it won't collapse back to 1 line only.

http://phpbbsocialnetwork.com/viewtopic.php?f=62&t=4183


Culprit-CZ:
Need much more effective solution

added to elastic options

``` js
parentElement
submitElement
```

When mouse is over `submitElement` than blur will not collapse.
`parentElement` is necessary to select the `submitElement`

Selector of the submitElement

``` js
$(this).parents(parentElement).find(submitElement)
```

I did not find any better solution for create simplest DOM selector

senky:
I think this is so serious codechange, that we should simply give up on it and mark it as purpose, that comment box does not collapse. This makes code extremely unreadable. Or at least this PR seems like it. Could you, please, write here what parts of code (for example state exact lines), which have been changed and which are just marked as changed by diff? Thanks!

Kamahl19:
Have you changed spaces for tabs again? I know it is necessary but maybe we should to it in separate PR, because this one is unreadable now

PR #58, opened 2012-10-09
Link to show your and user's posts is missing in Statistics tab in profile.php

senky:
In label "Total posts" after total number of your or user's posts, in common viewprofile page, there is a link titled with "Show your posts" if you are on your profile and "Search user's posts" if you are on other's profile. This is, however, missing in profile.php. It should be added there.

http://phpbbsocialnetwork.com/viewtopic.php?f=62&t=4205

### DIFF
--- a/root/socialnet/profile.php
+++ b/root/socialnet/profile.php
@@ -545,6 +545,7 @@ if (!class_exists('socialnet_profile'))
 				'POSTS'				 => ($member['user_posts']) ? $member['user_posts'] : 0,
 				'WARNINGS'			 => isset($member['user_warnings']) ? $member['user_warnings'] : 0,
 				'S_WARNINGS'		 => ($auth->acl_getf_global('m_') || $auth->acl_get('m_warn')) ? true : false,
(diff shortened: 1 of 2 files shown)


Culprit-CZ:
Kamahl work, than he must

root/socialnet/profile.php
```diff
@@ -545,6 +545,7 @@ function tab_stats($user_id)
 				'POSTS'				 => ($member['user_posts']) ? $member['user_posts'] : 0,
 				'WARNINGS'			 => isset($member['user_warnings']) ? $member['user_warnings'] : 0,
 				'S_WARNINGS'		 => ($auth->acl_getf_global('m_') || $auth->acl_get('m_warn')) ? true : false,
+				'U_SEARCH_USER'		 => ($auth->acl_get('u_search')) ? append_sid("{$phpbb_root_path}search.$phpEx", "author_id=$user_id&amp;sr=posts") : '',
```
Kamahl19: There is U_SEARCH_USER in root/profile.php , we can use that.

Kamahl19:
Btw this is not a bug but improvement, because we removed the link for 0.7.0

Culprit-CZ:
MA

PR #59, opened 2012-10-10
_PHP_SELF -> phpBB variable

Culprit-CZ:
http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=1486


Kamahl19:
I really dont know why we have used PHP_SELF before.

MA

PR #62, opened 2012-10-11
JavaScript compiler && IE8 and lower error

Culprit-CZ:
JavaScript compiler report warning.
IE8 and lower may report error.
###### Wrong

``` js
var = {a:1,b:2,}
```
###### Correct

``` js
var = {a:1,b:2}
```
## 
#### `CZ`

Standardni problem Internet Exploreru.
**Chybne**

``` js
var = {a:1,b:2,}
```

**Spravne**

``` js
var = {a:1,b:2}
```


senky:
MA

PR #51, opened 2012-10-11
Change twitter protocol to current one.

Kamahl19:
Twitter used this format of URL:
`http://twitter.com/#!/<nickname>/`
however it changed to:
`http://twitter.com/<nickname>/`

Our language key `'WRONG_DATA_TWITTER'` should reflect that change.

http://phpbbsocialnetwork.com/viewtopic.php?f=62&t=4187


Culprit-CZ:
MA

PR #67, opened 2012-10-15
Update jquery and jquery UI

Culprit-CZ:
[jQuery 1.8.2](http://blog.jquery.com/2012/09/20/jquery-1-8-2-released/) released
[jQuery UI 1.9](http://blog.jqueryui.com/2012/10/jquery-ui-1-9-0/) released


senky:
Updating vendors should make no harm to code itself. We should check if vendor's new code perserved compatibility with old version and if yes, no bug should appear at all. Besides, 0.x.0 will not be released so often, and I would not like to force addons to load theirs vendors just because we have outdated ones...

senky:
MA

PR #75, opened 2012-10-15
jquery-1.8.2.js should be named jquery-1.8.2.min.js

senky:
Due to file loading jQuery is called `jquery-1.8.2.js` and script loading it requests for `jquery-1.8.2.min.js`, this file should be renamed to `jquery-1.8.2.min.js`.


Kamahl19:
MA

Culprit-CZ:
What we did? Look at the code of the file

PR #78, opened 2012-10-15
removing permissions for statuses

Kamahl19:
it looks like we have not used that permission in any file. if you find it please let me know

### DIFF
--- a/root/socialnet/install_sn.php
+++ b/root/socialnet/install_sn.php
@@ -656,14 +656,20 @@ $versions = array(
 			'phpbbSN_smilies_allow',
 			'phpbbSN_create_fms_primarygroups',
 		),
-
-		'cache_purge'		 => array(
+	),
+	
+	'0.7.1'	 => array(
+	
+		'permission_remove'	 => array(
+			array('u_sn_userstatus', true),
+		),
+	
+  	'cache_purge'		 => array(
 			'imageset',
 			'template',
 			'theme',
 			'cache',
 		),
-
 	),
 );
 
--- a/root/language/en/acp/permissions_socialnet.php
+++ b/root/language/en/acp/permissions_socialnet.php
@@ -27,8 +27,6 @@ $lang['permission_cat']['socialnet'] = 'Social Network';
 // Adding the permissions
 $lang = array_merge($lang, array(
 	'acl_a_sn_settings'		 => array('lang' => 'Can edit Social Network settings', 'cat' => 'settings'),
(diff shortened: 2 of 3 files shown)


root/socialnet/install_sn.php
```diff
@@ -656,14 +656,20 @@
...
 			'phpbbSN_create_fms_primarygroups',
 		),
-
-		'cache_purge'		 => array(
+	),
+	
+	'0.7.1'	 => array(
+	
+		'permission_remove'	 => array(
+			array('u_sn_userstatus', true),
```
senky, line 664: Use tabs, please. Everywhere.

Kamahl19, line 664: I am using tabs

root/socialnet/install_sn.php
```diff
@@ -656,14 +656,20 @@
...
+		),
+	
+  	'cache_purge'		 => array(
```
senky, line 667: this line uses spaces - 2 spaces and 1 tab. It needs to be rewritten to use 2 tabs.

root/socialnet/install_sn.php
senky, line 664: Ok sorry, github displays it in strange way...

Culprit-CZ:
Why remove we permissions? What is the BUG?
**Permissions are used.**
When permission for module exist, than is used.

Better will be add permissions for all other modules

> u_sn_im
> u_sn_notify

I think that for FMS module, should be allways set permission to ALLOW.

senky:
In fact you are true. It maybe should not be removed, but for sure it needs to be set to "Yes" by default.

Kamahl19:
So what is your conclusion? Do you both want permissions or not? If yes, how can I discard changes from this PR ?

senky:
I think we should keep it. If it is not used in core of SN, this does not mean add-ons will not want to use it.

Culprit-CZ:
I'm for **discard** and **add permissions** for another modules and set it for group registered usesrs by default enabled

Kamahl19:
but what about other groups? it is useless and admins have to set it after installing. then they are asking why statuses are not working...

senky:
This issue is not about that. We should deal with permissions for statuses only. They should be set to "Yes" by default. That is it.

Kamahl19:
All right so how can I discard changes I made to remove it ?

senky:
Simply replace them by reverting changes. Or delete commits, it should revert the changes made in the commit.

PR #77, opened 2012-10-15
IM group bar expand/collapse image moved to left when bar above is collapsed

senky:
http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4232


Kamahl19:
MA

PR #81, opened 2012-10-15
Add prepare-commit-msg hook

senky:
To ease commit message creation, prepare-commit-msg hook should be added to `/git-tools/` so that it prepends commit message with `[<branch name>]` - what should usually be `i/<issue ID>`.


Culprit-CZ:
MA

Kamahl19:
MA

PR #82, opened 2012-10-16
SN theme

Culprit-CZ:
- Variables fix.
- Get message after compiling eg. compiled not compiled.
- Documentation of code


senky:
MA

Kamahl19:
MA

PR #85, opened 2012-10-16
Add .gitattributes to package

senky:
gitattributes file should be added to package to set some conventions. This file will force git to work the way we need.


Culprit-CZ:
MA

senky:
@Kamahl19 if you give MA to this, please do not merge it. As we agreed that I will squash my commits before it, please, just comment "MA" and leave mergin on me. Thanks!

Kamahl19:
MA

PR #74, opened 2012-10-16
Format date in profile.php Birthday according to user preferences

senky:
Currently, date in Birthday label in Info tab in profile.php uses force-written format. It should be up to user how to display it.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4243


Kamahl19:
So what is your conclusion?

senky:
We left decision on you.

Culprit-CZ:
We left decision on you.
1. You declare in lang format date for date of birth
2. We use regular expression to cut current part of user setting to display correctly date od birth

Kamahl19:
I am for 2nd solution

senky:
After long discussion we decided to use solution as proposed in the commit of this PR.

Culprit-CZ:
MA

PR #33, opened 2012-10-16
Load older posts earlier than user hits bottom of the page

senky:
Older posts should start loading earlier than user hits the very bottom of the page. It should start at least at 200px before bottom.


Kamahl19:
MA

woipi90:
i think the posts should be loaded even earlier..

senky:
The PR has already been marged, you can post a new topic to Feature requests forum if you think...

PR #64, opened 2012-10-17
No sound for new massage in certain conditions

Culprit-CZ:
http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4225

### DIFF
--- a/root/socialnet/im.php
+++ b/root/socialnet/im.php
@@ -539,6 +539,8 @@ if (!class_exists('socialnet_im'))
 
 			$b_no_avatar_me = stripos($this->config['my_avatar'], 'socialnet/no_avatar') !== false ? true : false;
 
+			$play_sound_on_load = false;
+			
 			for ($i = 0; isset($chatBoxRowSet[$i]); $i++)
 			{
 				$row = $chatBoxRowSet[$i];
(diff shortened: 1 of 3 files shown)


Culprit-CZ:
##### `m.im.js`

EOL CR to LF

Added lines
46

``` js
playSoundOnPageLoad: false
```

345-352

``` js
// Play sound
if ($.sn.im.opts.sound && $.sn.im.playSoundOnPageLoad) {
    if ($.browser.msie) {
        $('#sn-im-msgArrived').html('<object height="1" width="1" type="application/x-shockwave-flash" data="' + $.sn.im.opts.soundFile + '"><param name="movie" value="' + $.sn.im.opts.soundFile + '"><param name="FlashVars" value="' + $.sn.im.opts.soundFlashVars + '"></object>');
    } else {
        $('#sn-im-msgArrived').html('<embed src="' + $.sn.im.opts.soundFile + '" width="0" height="0" pluginspage="http://www.macromedia.com/go/getflashplayer" type="application/x-shockwave-flash" FlashVars="' + $.sn.im.opts.soundFlashVars + '"></embed>');
    }
}
```

root/socialnet/im.php
```diff
@@ -630,6 +636,8 @@ function _startIM($mode = '')
 
 				$this->_markRecievedMessages();
 			}
+			
+			$template->assign_var('SN_IM_PLAY_SOUND_ON_PAGELOAD', $play_sound_on_load?'true':'false');
```
senky: This should look like:

`$template->assign_var('SN_IM_PLAY_SOUND_ON_PAGELOAD', $play_sound_on_load ? 'true' : 'false');`
just decorating change, but we really should follow coding guidelines.

senky:
MA

PR #90, opened 2012-10-18
Correct logic in passing $format in snFunctions::date_format()

senky:
Currently we use

```
if ($format == false || $formatUser != '')
```

but it should be:

```
if ($format == false && $formatUser != '')
```


senky:
This also replaces CRLF to LF, so real diff is:

```
line 835
-               if ($format == false || $formatUser != '')
+               if ($format == false && $formatUser != '')

line 845
-               if ($format == false || $formatUser != '')
+               if ($format == false && $formatUser != '')
```

Kamahl19:
MA

Culprit-CZ:
MA

PR #80, opened 2012-10-18
Add "clear window" button to chatbox

senky:
Chatbox should have new feature included - a button that clears content of chatbox using JS.

http://phpbbsocialnetwork.com/viewtopic.php?f=6&t=4229


root/styles/prosilver/template/socialnet/im.html
```diff
@@ -53,7 +55,7 @@
 				<!-- BEGIN message -->
 				  <div class="sn-im-msg<!-- IF sn_im_chatbox.message.S_UID_SAME --> sn-im-noborderTop<!-- ENDIF --> {time:{sn_im_chatbox.message.TIME},from:{sn_im_chatbox.message.UID_FROM}}">
 				    <span class="sn-im-msgTime">{sn_im_chatbox.message.TIME_STRING}</span>
-				    <!-- IF $USE_AVATAR eq 0 --> 
+						<!-- IF $USE_AVATAR eq 0 -->
```
senky, line 58: spaces replaced by one tab

root/language/en/mods/socialnet.php
```diff
@@ -38,6 +38,7 @@
 	'SN_IM_LOGIN'							 => 'Online',
 	'SN_IM_LOGOUT'							 => 'Offline',
 	'SN_IM_PRESS_TO_CLOSE'					 => 'Press %1$s to close chat box',
+	'SN_IM_PRESS_TO_CLEAR_WINDOW'			 => 'Press this to clear the window of the chat box',
```
Kamahl19, line 41: Use just 'Clear the chat box',

Kamahl19:
MA

root/language/en/mods/socialnet.php
senky, line 41: Why? I think if we want to keep consistancy, we should display similar messages in each action...

Kamahl19, line 41: Agree, we should edit SN_IM_PRESS_TO_CLOSE too to 'Close chat box'. Why is there %s?

balboag:
I think new options should be in a dropdown button like facebook, to many icons...

Kamahl19:
+1

balboag:
Only two buttons: dropdown button (gear icon) and close button (x), then, when you click on dropdown button you see (text only): View history chat, Clear the chatbox, and ¿report user?

root/language/en/mods/socialnet.php
senky, line 41: This is, however, not part of this PR. Whatsmore, %1$s is here because user can close chat box with Esc or another key it is bind to it. We really should keep this wording.

senky:
There are just 3 buttons for now. They make less than 20% of overall width. I think it is not necessary for now. If new buttons will appear, I am not against it, but just try to install it - you will see it makes no harm to keep it like this.

senky:
@Kamahl19 please download it and test it to ensure it works as it should.

sn-theme/prosilver/classes_im.less
```diff
@@ -459,6 +459,13 @@
 	background-image:url('{T_IMAGESET_PATH}/socialnet/button_smiley.png');
 }
 
+.ui-icon.sn-im-cbClear {
+	background-image:url('{T_IMAGESET_PATH}/socialnet/button_clear.png');
+	width: 16px;
```
Culprit-CZ: Why? Already set. `.ui-icon` have this width

sn-theme/prosilver/classes_im.less
```diff
@@ -459,6 +459,13 @@
...
+	height: 16px;
```
Culprit-CZ: Why? Already set. `.ui-icon` have this height

root/socialnet/js/m.im.js
```diff
@@ -803,6 +832,24 @@
 			}
 		},
 
+		_clearWindow : function(object) { // object = $('.sn-im-cbClear')
+			// load time from cookie specific for currently logged in user and chatbox he/she is on
+
+			cookie_time = $.parseJSON($.cookie('chat_clear'))[$.sn.user_id]['' + object.parents('.sn-im-block').attr('id').replace('sn-im-chatBoxBlock', '')];
```
Culprit-CZ: Does not correspond to SN.  Use `$.sn.getCookie` or `$.sn.setCookie` instead `$.cookie`.
Than is possible catch the cookie trough phpBB.

When cookie is not set than

```
TypeError: $.parseJSON($.cookie("chat_clear")) is null
http://dev.phpbb3hacks.com/sn_culprit/root/socialnet/js/m.im.js
Line 729
```

root/socialnet/js/m.im.js
```diff
@@ -338,12 +338,41 @@
...
+
 			if ($('.sn-im-block .sn-im-msgs:visible').is(':visible')) {
 				var $block = $('.sn-im-block .sn-im-msgs:visible').parents('.sn-im-block');
 				this._cwClose($block);
 				this._cwOpen($block, false);
 			}
 			this._startTimers();
+
+			// clear windows if needed
+			$('.sn-im-title .sn-im-cbClear').each(function() {
```
Culprit-CZ: Not necessary, could be filtered directly by PHP.
Cookie are visible in phpBB.

This should be placed by another events corresponding to IM title buttons

``` js
$('.sn-im-cbClear').each(function() {
```

is enough

root/socialnet/js/m.im.js
```diff
@@ -338,12 +338,41 @@
...
+			/** Clear window of the chat box */
+			$('.sn-im-cbClear').live('click', function() {
+				// we fetch user id from id attribute of chatbox parent element
+				var user_id = $(this).parents('.sn-im-block').attr('id').replace('sn-im-chatBoxBlock', '');
+
+				// unix timestamp
+				var time = Math.round((new Date()).getTime() / 1000);
+
+				// object of all clears for this browser and user
+				var clears = $.parseJSON($.cookie('chat_clear')) || {};
```
Culprit-CZ: Does not correspond to SN.  Use `$.sn.getCookie` or `$.sn.setCookie` instead `$.cookie`.
Than is possible catch the cookie trough phpBB.

Cookie name does not correspond to SN name conventions.

root/socialnet/js/m.im.js
```diff
@@ -338,12 +338,41 @@
...
+				// clear window!
+				$.sn.im._clearWindow($(this));
+			});
+
 			if ($('.sn-im-block .sn-im-msgs:visible').is(':visible')) {
 				var $block = $('.sn-im-block .sn-im-msgs:visible').parents('.sn-im-block');
 				this._cwClose($block);
 				this._cwOpen($block, false);
 			}
 			this._startTimers();
```
Culprit-CZ: This line should be the last line of the initialize function

Culprit-CZ:
##### `root/styles/prosilver/imageset/socialnet/button_clear.png`

This image does not correspond to our images, use this one
http://dev.phpbb3hacks.com/sn_culprit/blank.png

root/socialnet/js/m.im.js
senky: It is not enough, because `.sn-im-cbClear` is included 2 times - in `.sn-im-title` and `.sn-im-buttons`. If I do not filter it by `.sn-im-title`, all clears are trigerred 2 times.

root/socialnet/js/m.im.js
senky: What should the cookie name be? `chat-clear`?

Culprit-CZ:
MA

Kamahl19:
MA

PR #100, opened 2012-10-21
Minimum time to check for new IMs option in ACP is set to 1 by default

senky:
Minimum time to check for new IMs option in ACP is set to 1 by default, while it can be set to 2 at minimum:

```
The provided value for the setting “Minimum” is too low. The minimum acceptable value is 2.
```

This error is displayed even before you submit form.

This option should be set to `2` by default and checked in update script to be updated to `2` if it is set to `1`.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4290

### DIFF
--- a/git-tools/update_database_dev.php
+++ b/git-tools/update_database_dev.php
@@ -667,8 +667,14 @@ $versions = array(
 		),
 
 	),
-	
+
 	'0.7.1'	 => array(
+	),
+
+	'0.7.2-dev1'	=> array(
+		'custom'			 => array(
+			'phpbbSN_sanatize_im_checkTime_min',
+		),
 	),
 );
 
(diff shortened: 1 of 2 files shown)


git-tools/update_database_dev.php
```diff
@@ -678,6 +684,37 @@ function show_usage()
...
+				FROM ' . SN_CONFIG_TABLE . '
+				WHERE config_name = "im_checkTime_min"';
+		$result = $db->sql_query($sql);
+		$im_checkTime_min = $db->sql_fetchfield('config_value');
+
+		if ($im_checkTime_min == 1)
+		{
+			$db->sql_query('UPDATE ' . SN_CONFIG_TABLE . ' SET config_value = "2" WHERE config_name = "im_checkTime_min"');
+
+			return 'Social Network::IM check time minimum was set to 2';
```
Kamahl19, line 706: I dont think we need to return a notice about this.

Kamahl19:
MA but I think we should remove return messages

senky:
Why? It is used in any other custom functions in UMIL. Admin should be noticed that something has been changed in his board without asking him before.

Kamahl19:
allright

git-tools/update_database_dev.php
Culprit-CZ, line 706: UMIL returns is necessary. We should give some info to output.

senky:
d4ef4db merged this PR.

PR #104, opened 2012-10-22
Default IM min checkTime

Culprit-CZ:
http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4290


Culprit-CZ:
Someone was faster #100

PR #103, opened 2012-10-22
Addon Hook System does not count with inherited templates

Culprit-CZ:
If you use inherited template, hook system for addons does not count with it, so it tries to load template files from currently used style, but it should load template files from parent's style.

The problem is at least in sn_core_addons.php at [a line 120](https://github.com/phpBB-Social-Network/phpBB-Social-Network/blob/develop/root/socialnet/includes/sn_core_addons.php#L120).

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4257

### DIFF
--- a/root/socialnet/includes/sn_core_addons.php
+++ b/root/socialnet/includes/sn_core_addons.php
@@ -117,11 +117,6 @@ class sn_core_addons
 			}
 			$addonTemplate = $this->get_template_name($addon['addon_php'], $addon['addon_function'], $addon['ph_script'], $addon['ph_block']);
 
-			if (!file_exists("{$phpbb_root_path}styles/{$user->theme['template_path']}/template/socialnet/addons/{$addonTemplate}"))
-			{
-				continue;
-			}
-
 			include_once("{$phpbb_root_path}socialnet/addons/{$addon['addon_php']}.{$phpEx}");
 
 			$addonClass = new $addon['addon_php']($this->p_master);


root/socialnet/includes/sn_core_addons.php
```diff
@@ -50,8 +50,6 @@ class sn_core_addons
 	 */
 	public function sn_core_addons(&$p_master)
 	{
-		global $cache, $template, $phpbb_root_path, $user, $config;
```
senky: Why is this removed in this PR? We agreed that we will create separated PRs for it.

root/socialnet/includes/sn_core_addons.php
```diff
@@ -117,11 +115,6 @@ public function get($script = null, $block = null)
 			}
 			$addonTemplate = $this->get_template_name($addon['addon_php'], $addon['addon_function'], $addon['ph_script'], $addon['ph_block']);
 
-			if (!file_exists("{$phpbb_root_path}styles/{$user->theme['template_path']}/template/socialnet/addons/{$addonTemplate}"))
```
senky, line 120: Removing this condition will fix the bug? I just do not understand why it was added here.

root/socialnet/includes/sn_core_addons.php
Culprit-CZ: Sure but.
Why not yet? File is committed and is correct

root/socialnet/includes/sn_core_addons.php
Culprit-CZ, line 120: Yes, this will fix the problem.

root/socialnet/includes/sn_core_addons.php
senky: It includes something that should not be here. Afterwards we will not have properly sorted commits, what is the basic idea why we push stuff using PR.

root/socialnet/includes/sn_core_addons.php
senky, line 120: Are you sure it does not reproduce another problem? Why it was added here? What should it prevent from?

Culprit-CZ, line 120: Stop this discuss.
It check only existing file template. If the file does not exist then occur phpbb template error. This removed line avoid the problem for non existing template. So now is removed and when no assigned template exists then occur standard phpBB template error.

Kamahl19:
MA

senky:
Ok, just return `global` back ( https://github.com/phpBB-Social-Network/phpBB-Social-Network/pull/103/files#r1907931 ) and I will give you MA.

root/socialnet/includes/sn_core_addons.php
```diff
@@ -51,7 +51,7 @@ class sn_core_addons
 	public function sn_core_addons(&$p_master)
 	{
 		global $cache, $template, $phpbb_root_path, $user, $config;
-
+
```
senky: Remove blank spaces, please.

PR #97, opened 2012-10-22
Improve langs for 0.7.1

Kamahl19:
Fix some bugs according this diff http://phpbbsocialnetwork.com/viewtopic.php?f=25&t=4275&p=18148#p18148


Kamahl19:
i dont know why but it added the first commit. i was trying to fix my problem with git according culprits instructions but it seems it was not good solution

senky:
Alright, we won't avodit that - since adding .gitattributes, all opened files are forced to be changed to LF EOL. You cannot avoid that, nor revert it back. We will simply merge it as it is.

MA

senky:
Please, remove ca8ee98 so that we can merge it.

Culprit-CZ:
MA

PR #96, opened 2012-10-23
Time for all activities

Kamahl19:
We should unify activities and statuses and add time to activities too


Kamahl19:
It happend again. How can I delete this commit from my PRs ?

Kamahl19:
Problem is with the "solving the problem with git" commit. How can I remove it ?

Culprit-CZ:
MA

senky:
MA.

And please just wait with merging this PR untill we will release 0.7.1 (and 0.7.2 if needed). If we merge it now, it will be included in 0.7.1, too, what is undesired.

PR #106, opened 2012-10-23
Limitting HOOK calls

Culprit-CZ:
1. Limitting HOOK calls
2. Less DB queries
3. Less time load.


Kamahl19:
MA

senky:
The first commit (f46ee1c) lacks prefix. Otherwise it looks ok.

PR #111, opened 2012-10-24
Limitting HOOK calls

Culprit-CZ:
#106 no code attached

### DIFF
--- a/root/includes/hooks/hook_socialnet.php
+++ b/root/includes/hooks/hook_socialnet.php
@@ -43,8 +43,6 @@ class hookSocialNet
 	{
 		global $db, $user, $socialnet, $config, $template, $phpbb_hook;
 
-		$phpbb_hook->remove_hook('phpbb_user_session_handler', array('hookSocialNet', 'start_socialNet'));
-		
 		$sql = "SELECT config_value FROM " . SN_CONFIG_TABLE . " WHERE config_name = 'sn_global_enable'";
 		$result = $db->sql_query($sql);
 		$row = $db->sql_fetchrow($result);
--- a/root/socialnet/includes/socialnet.php
+++ b/root/socialnet/includes/socialnet.php
@@ -60,6 +60,7 @@ class socialnet extends snFunctions
 		"10",
 	);
 
+	var $useTemplateHook = true;
 	var $config = array();
 	var $socialnet_root_path = '';
 	var $modules = array();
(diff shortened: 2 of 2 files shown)


root/includes/hooks/hook_socialnet.php
```diff
@@ -43,8 +43,6 @@ static function start_socialNet()
 	{
 		global $db, $user, $socialnet, $config, $template, $phpbb_hook;
 
-		$phpbb_hook->remove_hook('phpbb_user_session_handler', array('hookSocialNet', 'start_socialNet'));
```
senky, line 46: Can this be removed? The code in `root/socialnet/includes/socialnet.php` prevents running it repeatedly?

Culprit-CZ, line 46: But remove the hook is by first call. Than should be removed

senky, line 46: Alright. I do not understand this part of code, yet. So if this is proper handling, then I am OK with the proposed code.

Culprit-CZ, line 46: phpBB hook is called by `$template->display` or `$template->assign_display`
When we call in our hook `$phpbb_hook->remove_hook` than is **not** the hook called by the next  `$template->display` or `$template->assign_display`
So we added `$socialnet->useTemplateHook` to call whole hook only once. Than we no longer need the phpbb hook removing.

senky, line 46: Alright, thanks for your explanation.

root/socialnet/includes/socialnet.php
```diff
@@ -695,7 +696,10 @@ function page_header($page_title = '', $display_online_list = true, $item_id = 0
 	function page_footer($block = 'body', $print = false)
 	{
 		global $template;
+		
+		$this->useTemplateHook = false;
```
senky, line 700: I think this is the way how it is possible to run hook twice, does not it? Because you set it to `false`, then run `$template->assign_display` what runs hook second time, as `$this->useTemplateHook` is set to `false` for now.

Culprit-CZ, line 700: The hook is terminated at start when  `$this->useTemplateHook` is set to  `false`

senky, line 700: Ah sorry, thanks for your explanation.

senky:
MA

Kamahl19:
MA

PR #113, opened 2012-10-31
Textbox overflow on copy/paste of bigger text 

Culprit-CZ:
It is problem with `.sn-us-mentions`. It's height needs to be updated along with updating height of the textarea.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4307

### DIFF
--- a/sn-theme/prosilver/classes_userstatus.less
+++ b/sn-theme/prosilver/classes_userstatus.less
@@ -176,11 +176,11 @@
   }
 
   .sn-us-mentions{
-  	top: 1;
-
-		line-height: 18px;
+	line-height: 18px;
     color: #fff;
     position: absolute;
+	top: -999px;
+	left: -999px;
     overflow: hidden;
     white-space: pre-wrap;
     word-wrap: break-word;
@@ -192,7 +192,6 @@
 		.rtl & {
 			right: 1;
 		}
-
 
     & > div{
       color: #fff;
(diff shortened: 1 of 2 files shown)


senky:
Along with `sn-theme/prosilver/classes_userstatus.less`, `root/styles/prosilver/theme/socialnet.css` must be updated and sent to this PR.

sn-theme/prosilver/classes_userstatus.less
```diff
@@ -178,9 +178,11 @@
   .sn-us-mentions{
   	top: 1;
```
senky: As you state `top` attribute in line 184, you can delete this one.

senky:
I have tested it, and it works, however I quite do not understand why. Can you explain me, how can negative `top` and `left` cause, that text does not overflows any more?

Culprit-CZ:
How to explain?
Elastic (expanding textarea) plugin create a `div` that is placed before.
The position of this `div` was same as position for the `textarea`.
For right count of the height should be not `hidden`.

sn-theme/prosilver/classes_userstatus.less
```diff
@@ -198,6 +198,7 @@
       color: #fff;
       white-space: pre-wrap;
       width: 100%;
+
```
senky: This line should not be here, I believe.

senky:
Thanks for your explanation! I like you solved it without changing Javascript :)

sn-theme/prosilver/classes_userstatus.less
Culprit-CZ: It is irrelevant, It is our dev code.
No problem I will remove.

Culprit-CZ:
I'm glad that it could be solved without JavaScript

senky:
MA

Culprit-CZ:
@Senky please control my squash

senky:
Perfect!

MA

PR #84, opened 2012-10-16
Remove phpBB hook after hook function is called

Culprit-CZ:
#### Error

```
Fatal error: Allowed memory size of 44040192
bytes exhausted (tried to allocate 16384 bytes) in
/root/includes/template.php on line 242
```
#### Causes

phpBB hook is called recursively

PR #112, opened 2012-10-31
$phpbb_hook->remove_hook should not be used

Culprit-CZ:
[URLs are not replacing for FMS ➽](http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4281&p=18457#p18453)
associated with #111, #106 and #84.

**Two ways:**
1. replace template hook functions for replace HREF in HTML code
2. Find some way how to solve correctly calling template hook for each parts for `$template->assign_display`


Culprit-CZ:
http://phpbbsocialnetwork.com/viewtopic.php?f=21&t=4354 related to this fix

senky:
I think commit message says exactly opposite of what it changes. You are adding session handler hook and removing template hook.

However when we remove template hook, I think that links that have been parsed from memberlist to profile.php will not be parsed any more. Or am I wrong?

Culprit-CZ:
I thought I was correct, as you say, but it sounded reversed

**If we remove any hook, it will remove all hooks, which they are assigned.
Therefore it is not possible to remove hook.**

root/includes/hooks/hook_socialnet.php
```diff
@@ -125,11 +127,8 @@ static function template_display($phpbb_hook, $handle, $include_once = true)
...
 }
-
 /**
  * Register all necessary hooks
  */
 $phpbb_hook->register('phpbb_user_session_handler', array('hookSocialNet', 'start_socialNet'));
 $phpbb_hook->register(array('template', 'display'), array('hookSocialNet', 'template_display'));
-
-?>
\ No newline at end of file
```
senky: There should be a new line at the end of file.

senky:
That is why I strongly encouraged you to use singleton pattern to sn-hook class. You use constant, what is fine, however I would likely use private variable...

Please, add new line at the end of file (and optionally change constant to private variable) and I will give you MA.

root/socialnet/notify.php
```diff
@@ -402,92 +402,6 @@ function ntf_notify_count($status = SN_NTF_STATUS_UNREAD)
...
-			{
-				print '<pre>';
-				print_r($template);
-				die(__FILE__ . ' ' . __LINE__);
-			}
-		}
-	}
-}
-
-if (!function_exists('hook_template_confirmBox_URL_array_callback'))
```
senky, line 476: You deleted this function. But why? We do not need it anymore?

senky:
MA

PR #118, opened 2012-11-03
Inactive users are shown in "Users you may know" block

Culprit-CZ:
User A register a new account but don't click the link in the mail, so he is inactive and not shown in memberlist.

User B is registered user and sn shows User A on activity page block "you may know".

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4373

### DIFF
--- a/root/socialnet/includes/functions.php
+++ b/root/socialnet/includes/functions.php
@@ -1332,7 +1332,10 @@ class snFunctions
 
 			$sql = "SELECT DISTINCT u.user_id, u.username, u.user_avatar, u.user_avatar_type, u.user_avatar_width, u.user_avatar_height, u.user_colour
 					FROM " . ZEBRA_TABLE . " AS z, " . USERS_TABLE . " AS u
-					WHERE " . $db->sql_in_set('z.user_id', $ap_my_friends_ary, false, true) . " AND " . $db->sql_in_set('z.zebra_id', $ap_my_friends_me_ary, true, true) . " AND z.friend = 1 AND z.zebra_id = u.user_id";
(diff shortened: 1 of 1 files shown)


root/socialnet/includes/functions.php
```diff
@@ -1332,7 +1332,10 @@ function friends_suggestions($limit = 4)
 
 			$sql = "SELECT DISTINCT u.user_id, u.username, u.user_avatar, u.user_avatar_type, u.user_avatar_width, u.user_avatar_height, u.user_colour
 					FROM " . ZEBRA_TABLE . " AS z, " . USERS_TABLE . " AS u
-					WHERE " . $db->sql_in_set('z.user_id', $ap_my_friends_ary, false, true) . " AND " . $db->sql_in_set('z.zebra_id', $ap_my_friends_me_ary, true, true) . " AND z.friend = 1 AND z.zebra_id = u.user_id";
+					WHERE " . $db->sql_in_set('z.user_id', $ap_my_friends_ary, false, true) . "
+						AND " . $db->sql_in_set('z.zebra_id', $ap_my_friends_me_ary, true, true) . "
+						AND z.friend = 1 AND z.zebra_id = u.user_id
+						AND u.user_type IN (" . USER_NORMAL . "," . USER_FOUNDER . ")";
```
senky: No - it should be defined like:

``` php
AND u.user_type NOT IN (" . USER_INACTIVE . "," . USER_IGNORE . ")";
```

Another MODs can add new user types, you cannot ignore this fact.

Culprit-CZ: You have right

Kamahl19:
MA

senky:
MA

PR #119, opened 2012-11-03
Update copyright message

senky:
I think we should update copyright to:

Powered by phpBB Social Network 0.7.0 © phpBB SN Group 2010-2012

or something like that

http://phpbbsocialnetwork.com/viewtopic.php?f=31&t=4370


Culprit-CZ:
MA

Kamahl19:
MA

PR #121, opened 2012-11-05
Search user’s posts does not work in Statistics tab

senky:
Search user’s posts does not work in users profile statistic tab.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4368


Culprit-CZ:
MA

PR #116, opened 2012-10-30
JS Social Network own namespace declarations

Culprit-CZ:
Javascript declaration changed from extending jQuery to own namespace `socialNetwork`

These files are assigned to #80 and #108.
#80 contains also file `m.socialnet.js`, but the change is only by adding one property.

This property is also added to this file in this PR.

PR #124, opened 2012-11-09
Comment boxes of more loaded statuses does not resize on focus

Culprit-CZ:
If you click on comment box, it doubles its height. However, when you load more posts, comment boxes of those loaded statuses does not double its height on focus.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4415

### DIFF
--- a/root/socialnet/js/jquery.elastic.source.js
+++ b/root/socialnet/js/jquery.elastic.source.js
@@ -46,7 +46,7 @@
 				}
 
 				// Elastic only works on non initialized objects
-				if (jQuery(this).attr('data-elastic') === 'elastic') {
+				if (this.dataElastic === 'elastic') {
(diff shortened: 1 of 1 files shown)


Culprit-CZ:
It is now independent from #116

Kamahl19:
MA

senky:
I just realised that it was using string instead of boolean because it used data-elastic properity, which is used by .elastic plugin, too. So it can be reverted back to 31b6055 and use string comparison in condition if you like (and it seems to be better solution, too, because it is compatibile with .elastic plugin).

Culprit-CZ:
Done

senky:
MA

Please, reword commit so that it is more descriptive (just copy title of this PR) and you can merge it.

PR #126, opened 2012-11-15
Automatic load of older posts on profile does not work

senky:
Automatic load of older posts on profile does not work. It is because in https://github.com/phpBB-Social-Network/phpBB-Social-Network/blob/develop/root/socialnet/js/m.userstatus.js#L570 there is

``` js
$('.sn-ap-getMore').size()
```

what should be rewritten to

``` js
$('.sn-us-getMore').size()
```

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4421


Culprit-CZ:
MA

Kamahl19:
MA

PR #129, opened 2012-11-19
jQuery UI 1.9.1 + jQuery 1.8.3

Culprit-CZ:
I think with some merge we lose this change.

http://phpbbsocialnetwork.com/viewtopic.php?f=18&t=4489


senky:
MA

Kamahl19:
MA

PR #132, opened 2012-11-25
Profile field edit using textarea does not parse HTML entities.

senky:
Profile fields which use textearea to be edited does not parse HTML entities. It is due to using `.text` instead of `.html`.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4451


Kamahl19:
MA

Culprit-CZ:
MA

PR #134, opened 2012-11-26
CSS problem with overlapping text with icon

Culprit-CZ:
http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4417

### DIFF
--- a/sn-theme/prosilver/classes_userstatus.less
+++ b/sn-theme/prosilver/classes_userstatus.less
@@ -23,6 +23,7 @@
 }
 
 .sn-us-buttonCommentOver {
+	visibility: hidden;
 	display: none;
 }
 
@@ -99,6 +100,7 @@
 }
 
 .sn-us-fetchPreview, .sn-us-thumbs {
+	visibility: hidden;
 	display: none;
 	padding: 4px 0;
 }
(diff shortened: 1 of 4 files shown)


Kamahl19:
MA

sn-theme/prosilver/classes_userstatus.less
```diff
@@ -193,10 +193,12 @@
 			right: 1;
 		}
 
+
```
senky: unnecessary line

sn-theme/prosilver/classes_userstatus.less
```diff
@@ -193,10 +193,12 @@
...
     & > div{
       color: #fff;
       white-space: pre-wrap;
       width: 100%;
+
```
senky: unnecessary line

senky:
PR lacks padding for ltr.

Plus, we agreed on using `visibility:hidden;` instead of `display: none` so that padding works also for hidden "X".

Culprit-CZ:
But `:not` is not a part of this PR.
Otherwise `:not` is not suporting only IE8

senky:
It is part of this PR:
https://github.com/phpBB-Social-Network/phpBB-Social-Network/pull/134/files#L0R947

What is the fallback for IE8? It is still commonly user browser, if `:not` is not supported there, how that parts of page will be displayed correctly?

Culprit-CZ:
Is it not, It's only your comments, that put it in this PR. Otherweise is it not part of this PR.
Did you tested?
We have tested on all browsers. Also .....

senky:
I just do not understand. I clicked Files changed tab and looked at the code. In root/styles/prosilver/theme/socialnet.css on line 947 I can see **new** code added using `:not` selector. What Is what my comment put into this PR?

Culprit-CZ:
File `root/styles/prosilver/theme/socialnet.css` changes are irelevant, because this file is each time fully compiled from less files. And as you see, in less files is not added `:not`.

senky:
Well, this is not the way I like it, nor I imagined the solution, but it works and fixes the bug, so:

MA

Kamahl19:
MA

PR #135, opened 2012-11-26
Allow zoom for mobile devices

Culprit-CZ:
When chat is disabled for lower resolution, than we could allow zoomable site.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4490


Kamahl19:
MA

PR #128, opened 2012-12-10

Kamahl19:

something like this

``` php
if (!function_exists('http_build_query')) {
function http_build_query($data, $prefix='', $sep='', $key='') {
   $ret = array();
   foreach ((array)$data as $k => $v) {
       if (is_int($k) && $prefix != null) $k = urlencode($prefix . $k);
       if (!empty($key)) $k = $key.'['.urlencode($k).']';

       if (is_array($v) || is_object($v))
           array_push($ret, http_build_query($v, '', $sep, $k));
       else    array_push($ret, $k.'='.urlencode($v));
   }

   if (empty($sep)) $sep = ini_get('arg_separator.output');
   return implode($sep, $ret);
}}
```


Culprit-CZ:
remove the https://github.com/Kamahl19/phpBB-Social-Network/commit/55896d18f20dc04aa6623df5a7ccc62074545966 and than I could give you MA

Culprit-CZ:
MA

senky:
`http_build_query()` is also used in [/root/socialnet/profile.php](https://github.com/phpBB-Social-Network/phpBB-Social-Network/blob/develop/root/socialnet/profile.php#L58), so it should be included in `snFunctions`.

Also I am not sure function you used supports array parametres, please, verify it.

Kamahl19:
good point

senky:
That is why I posted 2 versions of this function by myself. Second one should be capable of array parametres.

Also, fallback function must accept all parametres just like [original one](http://php.net/manual/en/function.http-build-query.php).

PR #139, opened 2012-12-11
Make HTML code more clear

senky:
1. root/styles/prosilver/template/socialnet/user_profile_body.html line 55 - `style="cursor:default;"` should be replaced by class.
2. root/styles/prosilver/template/socialnet/ucp_profile.html - all `<textarea>` elements must have `cols` attribute specified.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4542


senky:
To follow W3C, XHTML validation rules, remove browser errors, etc.

Kamahl19:
MA

Culprit-CZ:
MA

PR #140, opened 2012-12-12
ALT-tag in avatars in IM should not overflow

senky:
If image source of avatar in IM is not loaded, alt tag is used instead. It's content, however, overflows. It should be added `word-wrap: break-word;`.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4475


Kamahl19:
MA

Culprit-CZ:
MA

PR #125, opened 2012-12-12
Status and comment textarea does not resize back when it is empty and you click on Share/Comment button

senky:
When you focus status/comment textarea, do not write anything and then click on Share/Comment button, button is hidden, but textarea remains doubled height.

Button should not be hidden untill form is really submitted (simply - if you do not write anything, you cannot submit the form. Thus, whole form should be displayed untill you successfully submit it).

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4416

### DIFF
--- a/root/socialnet/js/m.activitypage.js
+++ b/root/socialnet/js/m.activitypage.js
@@ -5,7 +5,7 @@
  */
 
 /**
- * 
+ *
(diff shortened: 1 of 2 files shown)


Culprit-CZ:
MA

Kamahl19:
MA

PR #137, opened 2012-12-12
"You may know" block doesn't filter users who have been requested already

senky:
"You may know" block should not display users who have been requested already.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4419


Culprit-CZ:
MA

Kamahl19:
MA

PR #133, opened 2012-12-12
Blocked and deactivated users are shown in friends sections

senky:
If you block or deactivate user via ACP, it is still displayed in friends section all over the SN. SN friends listing function should contain automatic filter to remove blocked and deactivated users from list, so that we does not need to specify it manually in all friends listings.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4474

### DIFF
--- a/root/profile.php
+++ b/root/profile.php
@@ -141,10 +141,12 @@ if ($config['load_onlinetrack'])
 }
 
 // Load number of friends
-$sql = 'SELECT COUNT(zebra_id) AS num_friends
-    				  FROM ' . ZEBRA_TABLE . '
-    				    WHERE user_id = ' . $user_id . '
-                  AND friend = 1';
+$sql = 'SELECT COUNT(z.zebra_id) AS num_friends
+		FROM ' . ZEBRA_TABLE . ' z, ' . USERS_TABLE . ' u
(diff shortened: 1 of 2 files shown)


senky:
@Kamahl19: filter should be included in sql query. This means, it will filter out only users that are currently disabled or banned.

Kamahl19:
I dont know .. In my opinion we should not hide banned and deactivated users. I have no problem with them in friend lists. If user dont want them to appear in his friend list he can simply delete them. Otherwise he will wonder why some of his friends are not in his friendlist

senky:
The problem is, that you cannot communicate with that users in no way when they are deactivated/banned. You cannot chat with them, post statuses on their walls, nothing. They, in fact, do not exist. This is how it works in whole phpBB (just check memberlist).
But the thuth is, that banned users are displayed. However, it is another point of view to look at banned users in forums and in social network. I think banned users should be removed as well. I have attached commit that filters both. I can remove banned users filter, but I am against it. @Culprit-CZ, please, express yourself!

Culprit-CZ:
Totaly right, BANNED and IGNORED users should not be included.
MA

Kamahl19:
MA

PR #138, opened 2012-12-12
Notification count does not ignore deleted users

senky:
If you send notification by user A to user B and you delete user A before use B reviewed that notification, user B can see 1 unread notification, but is unable to see it after he opens notifications page.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4412


Culprit-CZ:
MA

Kamahl19:
MA

PR #109, opened 2012-12-13
Enable Friends managemend module after installation

senky:
http://phpbbsocialnetwork.com/viewtopic.php?f=2&t=4325


Culprit-CZ:
MA

Kamahl19:
MA

PR #143, opened 2012-12-14
Share buttons are not displayed with visibility

Culprit-CZ:
With adding `visibility: hidden` in #134 share buttons are not displayed on focus.


senky:
Please, remove all commits except from 0fc2ec5.

senky:
MA

Kamahl19:
MA

PR #144, opened 2012-12-14
Installation on SQLite does not work because of multiple primary key definitions in one table

senky:
Installation on SQLite failed, because of multiple primary key definitions in `<prefix>_sn_fms_groups` table. Second definition should be rewritten to UNIQUE keys.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=2120


Kamahl19:
MA

Culprit-CZ:
MA

PR #145, opened 2012-12-14
Include hook_sn-theme.php and .file_check to .gitignore

senky:
`root/includes/hooks/hook_sn-theme.php` and `sn-theme/*/.file_check` should be included in `.gitignore`.


Kamahl19:
MA

Culprit-CZ:
MA

PR #146, opened 2012-12-14
Friends with special characters can be dropped to friends group more times

senky:
If I add an user (with a normal username) in a group more times, he is visualized only a time. Instead, if I add an user (with special characters in username) in a group more times, he is visualized more times until page refresh.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4547


Kamahl19:
MA

Culprit-CZ:
Discussed.

MA

PR #147, opened 2012-12-14
functions in setTimeout have to use socialNetwork instead of $sn

senky:
`setTimeout` function calls function defined in the first parametre in global scoope. This means, we cannot use `$sn` shortcut.

### DIFF
--- a/root/socialnet/js/m.im.js
+++ b/root/socialnet/js/m.im.js
@@ -483,7 +483,7 @@
 			});
 
 		},
-		/** 
+		/**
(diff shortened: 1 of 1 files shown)


Kamahl19:
what errors was this causing?

senky:
Javascript errors, @Culprit-CZ knows more.

Kamahl19:
MA

Culprit-CZ:
`$sn` is not declared in main scope. There is `socialNetwork` declared. `$sn` is declared only in declarations and functions for shorter code

MA

PR #149, opened 2012-12-15
Replace PRIMARY keys in SN_FMS_GROUPS_TABLE by UNIQUE keys

senky:
As it was changed because of SQLite bug, update script should replace PRIMARY keys in old installations so that we do not have two different working installations.


Kamahl19:
MA

Culprit-CZ:
MA

PR #150, opened 2012-12-16
Replace readdir() & preg_match() with glob()

senky:
In `/root/socialnet/includes/socialnet.php` lines 28 - 36 combination of `opendir`, `readdir`, `preg_match` and `closedir` can be replaced with simple `glob()` (passes phpBB requirements) to make code simplier, clearer, and faster.

http://phpbbsocialnetwork.com/viewtopic.php?f=6&t=4615


Kamahl19:
MA

PR #151, opened 2012-12-16
Limit video fetch to youtube, metacafe and vimeo

senky:
Rich support of video providers in fetch feature causes lots of problems with managing and maintaining it. It should be limited to youtube, metacafe and vimeo.

http://phpbbsocialnetwork.com/viewtopic.php?f=6&t=4457

### DIFF
--- a/root/socialnet/includes/video.php
+++ b/root/socialnet/includes/video.php
@@ -18,9 +18,9 @@
  * @license   http://www.gnu.org/licenses/
  * @link      http://www.embevi.com
  * @package   EmbeVi
- * 
+ *
  */
- 
+
  /**
  * @ignore
  */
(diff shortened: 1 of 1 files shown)


Kamahl19:
I am against removing collegehumor.com (simmilar to vimeo, created by vimeo CEO), dailymotion.com and maybe also ustream.tv

If we remove all these services, then we dont need this class any more. It works with php5 so we could remove video.php and EmbeVi class and implement these few services to user status module.

Culprit-CZ:
I'm against to touching the `embeVi` class.
Upon the Kamahl comment we should remove whole script and implement by self, or we should leave `embeVi` class untouched.

MD

senky:
@Kamahl19 have you ever seen some of sites you proposed being emeded? Me not...

@Culprit-CZ if we do not touch it, it is probable, that still more and more sites will not fetch videos properly. We should either update it (but it looks like it is php5 only now), or create custom script, so that fetch function is not as consumptive as it is now (each fetch command needs to get through all of regexes defined in embeVi class).

PR #152, opened 2012-12-16
Do not hide notifications count even if it is 0

senky:
To keep consistency with phpBB and common rule (ie. Facebook), notifications count should not be hidden even if it's count is 0.


Kamahl19:
MA

Culprit-CZ:
MA
Upon the Senky's post t4624 merge I this PR

PR #153, opened 2012-12-16
Do not open internal links in a new window

senky:
Because it is just so logical!

http://phpbbsocialnetwork.com/viewtopic.php?f=6&t=4596


Kamahl19:
MA

Culprit-CZ:
MA for 0.7.2

PR #154, opened 2012-12-16
Tabify HTML

senky:
Template files are quite messy. It should be tabified properly so that it is easier to maintain.

http://phpbbsocialnetwork.com/viewtopic.php?f=6&t=4584


Kamahl19:
Have you used the Format function from some IDE or have you done it manually? Eclipse has great Format function

Kamahl19:
And I think we could include this to 0.7.3

Kamahl19:
MA

Culprit-CZ:
MA for 0.7.3

senky:
I did it myself so that it keeps logic. Anyway, I would not merge it untill it is properly tested. We should create test package and provide it to our testers, because there may be problems with whitespaces.

PR #156, opened 2012-12-20
Disable IM for mobile devices

Culprit-CZ:
We zoom enabled, but mobile devices have enabled the IM.
#135

But we want to disable IM for mobile devices.

### DIFF
--- a/root/socialnet/js/m.im.js
+++ b/root/socialnet/js/m.im.js
@@ -58,7 +58,7 @@
 				return false;
 			}
 
-			if ($sn.mobileBrowser && Math.min($(window).height(), $(window).width()) < 700 || $(window).width() < 650) {
+			if ($sn.mobileBrowser && Math.min(screen.availHeight, screen.availWidth) <= 720) {
 				$sn.enableModules.im = false;
 			}
 


Culprit-CZ:
I have tested on **iPad 2**, **iPhone 4**, **iPhone 5** and **Samsung Galaxy S III**

root/socialnet/js/m.im.js
```diff
@@ -58,7 +58,7 @@
 				return false;
 			}
 
-			if ($sn.mobileBrowser && Math.min($(window).height(), $(window).width()) < 700 || $(window).width() < 650) {
+			if ($sn.mobileBrowser && Math.min(screen.height, screen.width) < 750 || screen.width < 750) {
```
senky: I would use `availHeight` and `availWidth` as per http://www.w3schools.com/jsref/obj_screen.asp

And I would not change numbers, especially `Math.min(screen.height, screen.width) < 750`, because I have laptop with height of 738px (using screen.availHeight) what would result in not displaying chat, what is nonsense, because chatbox covers less than half of the height.

Culprit-CZ: Combination of `availHeight` and `availWidth` is necessary for mobile devices, because screen could be by mobile device flipped.

Culprit-CZ:
The combination of width and height is essential for mobile devices.
Mobile devices could be flipped.

Kamahl19:
I think we should merge it and then ask testers to use it with all cell phones they have

senky:
It is not about that. We do not want to disable it for all cell phones. We want to disable it for devices which have small screens. This is why the only deal here is to decide on number of pixels we will use. 750px seems to be too big, because as I wrote it already, my laptop screen is only 738px height, so it will not pass this condition (and yeah, I know it checks for dimensions only if device is detected as mobile browser, but to be honest, why chat should be hidden in iPad, when it has the same resolution as my laptop has?). Maybe 600px is better?

senky:
After private discussion:

MA

PR #157, opened 2012-12-20
Remove unused js files

Culprit-CZ:
I go trough the `overall_header_js.html` and find 3 js files that are not used.

So these files should be removed.


Kamahl19:
MA

PR #158, opened 2012-12-20

Kamahl19:

Automatically closed issue #128


Kamahl19:
I have created new issue and new branch, edited 2 files and sent PR but it added 3 more commits.. I have no idea why

Culprit-CZ:
1) You don't have fully updated branch develop from upstream before you create a new branch
2) Create a new branch with command `git branch --track <new branch name> upstream develop` and than checkout new branch

PR #159, opened 2012-12-21
CSS friend groups problem

Culprit-CZ:
t4629
http://phpbbsocialnetwork.com/viewtopic.php?t=4629&p=20329#p20278

margin right left problem


Kamahl19:
MA

PR #108, opened 2012-10-23
Smilies box for status

Kamahl19:
Display smilies box in status

PR #160, opened 2012-12-22

Kamahl19:

Automatically closed issue #158

### DIFF
--- a/root/socialnet/install_sn.php
+++ b/root/socialnet/install_sn.php
@@ -936,4 +936,49 @@ function phpbbSN_replace_primary_by_unique($action, $version)
 	return 'Updating of keys in SN_FMS_GROUPS_TABLE was successfully completed.';
 }
 
+if ( !function_exists('http_build_query'))
+{
+  /**
+   * Generate URL-encoded query string for php4
+	 * @see http://php.net/manual/en/function.http-build-query.php
+	 * @param array It may be a simple one-dimensional structure, or an array of arrays
+	 * @param string If numeric indices are used in the base array and this parameter is provided, it will be prepended to the numeric index for elements in the base array only
+	 * @param string Used to separate arguments
+	 * @param int Not used in function, just because of compatibility with php5 function
+	 * @return string Returns a URL-encoded string
+	 */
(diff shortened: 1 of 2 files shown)


senky:
Delete all commits but 56fdd7a. If you have problem with git, remove whole local repository and clone it again (do not forget to backup #108).

root/socialnet/install_sn.php
```diff
@@ -936,4 +936,40 @@ function phpbbSN_replace_primary_by_unique($action, $version)
 	return 'Updating of keys in SN_FMS_GROUPS_TABLE was successfully completed.';
 }
 
+if ( !function_exists('http_build_query'))
+{
+	function http_build_query($data, $prefix = '', $separator = arg_separator.output, $enc_type = PHP_QUERY_RFC1738)
```
senky: > The following constants are meant to be used with parse_url() and are available since PHP 5.1.2. 
> http://php.net/manual/en/url.constants.php

You cannot use `PHP_QUERY_RFC1738`.

root/socialnet/install_sn.php
```diff
@@ -936,4 +936,40 @@ function phpbbSN_replace_primary_by_unique($action, $version)
...
+					$queryString .= urlencode($correctKey) . "=" . urlencode($value) . $separator;
+				}
+				else
+				{
+					$queryString .= http_build_query($value, $correctKey, $separator, $enc_type);
+				}
+			}
+		}
+
+		return substr($queryString, 0, strlen($queryString) - 1);
```
senky: I think it should be:

``` php
return substr($queryString, 0, strlen($queryString) - strlen($separator));
```

senky:
MA

do not forget to squash the commits before the merge.

Culprit-CZ:
The function should be documented

Kamahl19:
It has the same documentation as the original php5 function

Culprit-CZ:
But it should be documented in code. When you look on the function should be known all parameters and from where is function taken, or some description.
We must document source code

Culprit-CZ:
MA

PR #162, opened 2012-12-27
Add proper formatting to birthdays in activitypage

senky:
Currently, date format in birthdays block in activitypage is force-formatted. It should use language-specific date format.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4517


Culprit-CZ:
MA

Kamahl19:
MA

PR #142, opened 2012-12-27
Do not add new line to new comment texarea on focus

senky:
Comment box should not resize when focused.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4543


senky:
This PR requires to change CSS, too, because it was (apparently) not properly compiled.

Culprit-CZ:
MA

**Remember: Before release will be published, must be CSS compiled again**

Kamahl19:
MA

PR #161, opened 2012-12-27
Weird search result for guests

senky:
When you are not logged in, you visit activitypage.php and type anything in the searchbox and submit it, it will redirect you to `socialnet/activitypage.php?mode=search` and content of this page is:

> {"user_id":"ANONYMOUS","more":false,"onlineCount":0}

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4649


woipi90:
and plz also include a fix for searching a name which includes "..

senky:
@weirdo90: what fix? Please, post it to the Bug tracker, so that we can review it properly.

Culprit-CZ:
MA

Kamahl19:
MA

PR #163, opened 2012-12-27
Underscore.js was removed, but it is used by mentionsInput

senky:
Underscore.js was removed, but it is used by mentionsInput. It should be included in the package again.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4654


Culprit-CZ:
My mistake, by locating all JS files.

MA

Kamahl19:
MA

PR #164, opened 2012-12-28
jQuery UI 1.9.2 update

Culprit-CZ:
New version of [jQuery UI](http://blog.jqueryui.com/2012/11/jquery-ui-1-9-2/). Released on Dec 23


Kamahl19:
MA

PR #165, opened 2012-12-28
It is possible to comment on statuses of non-friend user

senky:
`function get_friend_data` adds seeked user to friend list. It should not add user to friend list, nor be called "get_friend_data", because it gets user data, no matter if he/she is friend or not.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4653

### DIFF
--- a/root/socialnet/includes/socialnet.php
+++ b/root/socialnet/includes/socialnet.php
@@ -585,9 +585,9 @@ class socialnet extends snFunctions
 	/**
 	 * Get non-freinds data, no cache
 	 */
-	function get_friend_data($part, $user_id)
+	function get_user_data($part, $user_id)
 	{
-		global $user, $db;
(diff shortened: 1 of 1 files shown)


root/socialnet/includes/socialnet.php
```diff
@@ -585,7 +585,7 @@ function get_friend($mode, $user_id, $cfg_module, $cache_friends = true)
 	/**
 	 * Get non-freinds data, no cache
 	 */
-	function get_friend_data($part, $user_id)
+	function get_user_data($part, $user_id)
 	{
 		global $user, $db;
```
Culprit-CZ: `$user` is not necessary

Kamahl19:
MA

Culprit-CZ:
MA

PR #167, opened 2012-12-28
arg_separator.output is not always accessible

senky:
In our custom function `http_build_query`, we use `arg_separator.output`, but it is not always accessible. It should be replaced by string `'&'`.


Culprit-CZ:
MA

Kamahl19:
MA

PR #168, opened 2012-12-28
VIP: hooks

senky:
Introduce hooks to phpBB Social Network code.


root/socialnet/includes/sn_core_hook.php
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * $socialnet->sn_hook->add_action('<addon-identifier>.<tag>', 'custom_function_name');
+ *
+ *
+ * @package	phpBB-Social-Network
+ * @author 	Senky
+ * @version	0.7.2
+ * @access 	public
+ */
+class sn_hook {
```
Culprit-CZ, line 54: Why is not used `sn_core_hook` when is the name of the file `sn_core_hook.php`?

senky, line 54: It is just temporary name. To show you at least little bit of what I mean by hooks. It is not finished yet. Please, focus on logic.

PR #169, opened 2012-12-28
Smilies in chatbox are not displayed

senky:
Smilies in chatbox are loaded, but not displayed, because they have `visibility: hidden;`.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4661


Kamahl19:
MA

Culprit-CZ:
MA

PR #171, opened 2012-12-31
Reporting user is not properly styled for high resolution screens

senky:
"Reason" and "Further information" are next to each other in big resolution screens.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4678


Kamahl19:
MA

Culprit-CZ:
MA

PR #172, opened 2012-12-31
People you may know block displays people of non-confirmed friends

senky:
If you send an friend request to user, People you may know block displays his friends even if user has not approved your request yet.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4687


senky:
This one does not work. In fact, it never selects any user, because it is not possible to have `z.friend=1 AND z.approval = 1` true.

Kamahl19:
MA

Culprit-CZ:
MA

PR #173, opened 2012-12-31
Mutual friends are not displayed because of sql error

senky:
Mutual friends section in Friends tab in profile never displays mutual friends because of SQL error.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4677


Kamahl19:
MA

Culprit-CZ:
MA

PR #174, opened 2012-12-31
Suggestions user blocks does not have float: left

senky:
Suggestions block displays one user per row, because their wrapper has `float` set to `none`.
FireShot Screen Capture 102 - Social Network Activity page - dev_phpbb3hacks_com_socialnetwork_activitypage_php

http://phpbbsocialnetwork.com/viewtopic.php?p=20577#p20577


Kamahl19:
MA

Culprit-CZ:
MA

PR #175, opened 2012-12-31
Submitting empty status results in not collapsed textarea

senky:
If you submit empty status, you get notice dialog, textarea gets focused out, but does not collapse.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4694


Kamahl19:
MA

Culprit-CZ:
MA

PR #176, opened 2012-12-31
When deletting status, • should not be displayed

senky:
When deletting status, bull after time should not be displayed.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4692


Kamahl19:
MA

Culprit-CZ:
MA

PR #177, opened 2013-01-01
Reduce SN load

senky:
SN can be optimised in various ways to have less SQL queries.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4672


Culprit-CZ:
It seems to be ok
MA

Kamahl19:
MA

gooof:
no not work, a syntax error occurs.
SN_IM_PLAY_SOUND_ON_PAGELOAD shows only a value, no action performed by this.

PR #178, opened 2013-01-01
Optimise SN load, load only needed modules

senky:
Not all modules are needed in each page. Load only enabled and needed modules.
Reduce SQL queries, cache them, recycle them.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4684

### DIFF
--- a/root/profile.php
+++ b/root/profile.php
@@ -62,12 +62,30 @@ if ($user_id == ANONYMOUS && !$username)
 }
 
 // Select data for left column
-$sql = 'SELECT user_id, username, user_type, user_colour, user_inactive_reason, user_avatar, user_avatar_type, user_avatar_width, user_avatar_height, user_allow_pm
-              FROM ' . USERS_TABLE . '
-                WHERE ' . (($username) ? "username_clean = '" . $db->sql_escape(utf8_clean_string($username)) . "'" : "user_id = $user_id");
-$result = $db->sql_query($sql);
-$member = $db->sql_fetchrow($result);
-$db->sql_freeresult($result);
+if ( ($username && utf8_clean_string($username) == $user->data['username_clean']) || $user_id == $user->data['user_id'] )
+{
+	$member = array(
+		'user_id'	=> $user->data['user_id'],
+		'username'	=> $user->data['username'],
+		'user_type'	=> $user->data['user_type'],
+		'user_colour'	=> $user->data['user_colour'],
+		'user_inactive_reason'	=> $user->data['user_inactive_reason'],
+		'user_avatar'	=> $user->data['user_avatar'],
+		'user_avatar_type'	=> $user->data['user_avatar_type'],
+		'user_avatar_width'	=> $user->data['user_avatar_width'],
+		'user_avatar_height'	=> $user->data['user_avatar_height'],
+		'user_allow_pm'	=> $user->data['user_allow_pm'],
+	);
+}
+else
+{
+	$sql = 'SELECT user_id, username, user_type, user_colour, user_inactive_reason, user_avatar, user_avatar_type, user_avatar_width, user_avatar_height, user_allow_pm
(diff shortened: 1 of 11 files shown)


root/profile.php
```diff
@@ -62,12 +62,30 @@
 }
 
 // Select data for left column
-$sql = 'SELECT user_id, username, user_type, user_colour, user_inactive_reason, user_avatar, user_avatar_type, user_avatar_width, user_avatar_height, user_allow_pm
-              FROM ' . USERS_TABLE . '
-                WHERE ' . (($username) ? "username_clean = '" . $db->sql_escape(utf8_clean_string($username)) . "'" : "user_id = $user_id");
-$result = $db->sql_query($sql);
-$member = $db->sql_fetchrow($result);
-$db->sql_freeresult($result);
+if ( ($username && $db->sql_escape(utf8_clean_string($username)) == $user->data['username_clean']) || $user_id == $user->data['user_id'] )
```
gooof: $db->sql_escape(utf8_clean_string($username))
-> utf8_clean_string($username)

root/socialnet/im.php
```diff
@@ -122,7 +122,8 @@ function init()
 			$sendKey = $this->_keyboardString($this->sendSequence);
 
 			$sql = "SELECT * FROM " . SN_SMILIES_TABLE . " WHERE smiley_allowed = 1";
-			$rs = $db->sql_query($sql);
+			// cache for 1 year - smilies can be reloaded purging the cache
```
gooof: here you can try:
$cache->destroy('sql', SN_SMILIES_TABLE);
in ACP on every update/insert

senky: I think it is not needed. Users are used to purge cache in such situations, aren't they? It is like when you perform template updated - you need to clear the cache.

Whatsmore, `$cache->destroy` is really huge load-demanding function.

gooof: Ok correct. But is it not better than:
$cache->remove_file('sql_' . md5($cached) . '.' . $phpEx);
in socialnet/acp/acp_im.php? 
I see the remove_file() the first time.

I think the remove_file() is not working there, the SQL query will be trimmed, before the md5 will generate.
You need too add this line: (socialnet/acp/acp_im.php)
$cached = preg_replace('/[\n\r\s\t]+/', ' ', $cached); // from acm_file.php
before:
$cache->remove_file('sql_' . md5($cached) . '.' . $phpEx);

root/socialnet/acp/acp_im.php
```diff
@@ -137,12 +137,10 @@ function main_smilies($id)
 		if ($submit)
 		{
 			// PURGE CACHED SMILIES
-			$cached = 'SELECT ps.*, ss.smiley_allowed
```
senky, line 140: I could not find this query being used anywhere else, so this seems to be like dead code; I removed it and replaced it by needed clean-up.

Culprit-CZ:
MA

Kamahl19:
MA

root/socialnet/im.php
```diff
@@ -963,13 +964,40 @@ function _snImTyping()
...
-									WHERE user_id = '{$uidTo}'";
-			$rs = $db->sql_query($sql);
-			$rowAvatar = $db->sql_fetchrow($rs);
-			$db->sql_freeresult($rs);
+			global $db, $user, $phpbb_root_path, $chat_avatars;
+
+			$sql = 'SELECT user_avatar, user_avatar_type, user_avatar_width, user_avatar_height
+					FROM ' . USERS_TABLE . '
+					WHERE user_id = ' . (int) $uidTo;
+			$sql = preg_replace('/[\n\r\s\t]+/', ' ', $cached); // from acm_file.php
```
gooof, line 972: Undefined variable: cached

instead of $sql_md5 and preg_replace you can use here the $uidTo only

PR #179, opened 2013-01-01
It isn't possible to create friends group called "0"

senky:
You cannot create friends group called "0", even though it should be possible.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4696


Kamahl19:
MA

Culprit-CZ:
MA

PR #180, opened 2013-01-01
Contact info category in profile should be displayed only if it contains some content

senky:
Contact info category in profile should be displayed only if it contains some content.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4700


Kamahl19:
MA

Culprit-CZ:
MA

PR #181, opened 2013-01-01
Make sn-theme work like it was intended to

senky:
For now, sn-theme hook compiles `socialnet.css` after each reload of page. It was intended to compile `socialnet.css` only if `.less` files are changed.


Kamahl19:
MA

Culprit-CZ:
MA

PR #182, opened 2013-01-01
Add autofocus on username in issue2pr.php

senky:
Just to save time.


Kamahl19:
MA

Culprit-CZ:
I think it is necessary to give this page some design such as http://manage.phpbb3hacks.com/git-tools/#
Otherwise MA

senky:
Why do you need design for everything? It is simple, 4-input, internally used-only form. I can see no advantage in adding an design, in opposite to longer load and reindering of the page.

Culprit-CZ:
I like simple design. I hate/don't like designed pages.

senky:
So why do you you want to give a design to that page. The simpler it is, the better for us. It is not used by public, so we do not need to style it, IMHO.

senky:
After private discussion, Culprit is giving MA to this.

PR #186, opened 2013-01-02
Update jquery.detectmobilebrowser

Kamahl19:
Update jquery.detectmobilebrowser


senky:
`jquery.detectmobilebrowser.js` should be called `jquery.detectmobilebrowser.min.js`.

Culprit-CZ:
MA

PR #185, opened 2013-01-02
jGrowl should use phpBB css

Kamahl19:
- remove theme option from ACP
- use of phpbb template's css

### DIFF
--- a/git-tools/update_database_dev.php
+++ b/git-tools/update_database_dev.php
@@ -658,14 +658,6 @@ $versions = array(
 			'phpbbSN_smilies_allow',
 			'phpbbSN_create_fms_primarygroups',
 		),
-
-		'cache_purge'		 => array(
-			'imageset',
-			'template',
-			'theme',
-			'cache',
-		),
-
 	),
 
 	'0.7.1.dev1'	 => array(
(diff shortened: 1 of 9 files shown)


git-tools/update_database_dev.php
```diff
@@ -658,14 +658,6 @@ function show_usage()
 			'phpbbSN_smilies_allow',
 			'phpbbSN_create_fms_primarygroups',
 		),
-
```
senky, line 661: Why do you remove these lines?

Kamahl19, line 661: because cache_purge should always be only on the end of the file and not duplicated. I do it everytime I add new version to umil files

Culprit-CZ:
MA

root/socialnet/install_sn.php
```diff
@@ -656,14 +656,6 @@
 			'phpbbSN_create_fms_primarygroups',
 			'phpbb_SN_umil_send'
 		),
-
```
Culprit-CZ, line 659: Do not remove these lines from older install versions (659-666)

Culprit-CZ:
**`root/styles/prosilver/theme/jquery.jgrowl.css`**
I think it is original file from plugin. Why should we change it. Than we need to change file each time when jGrowl is upgraded.

Kamahl19:
That was the purpose of this issue, to change jgrowl css. Do you want to add css to SN css and just overwrite jgrowl css? 

If there will be new version of jgrowl, I dont think the CSS will be changed. And if yes, I have no problem in updating it.

root/socialnet/install_sn.php
Kamahl19, line 659: Do you mean cache_purge ? Why? They are at the end of the install script

Culprit-CZ:
I think that purpose of this issue is to use standard phpBB CSS and remove style option from ACP. Not to change original jQuery plugin jGrowl CSS file.

senky:
@Culprit-CZ, yes, it is about changing overall frontal look of jgrowl notifications. But @Kamahl19, what about removing original jgrowl file completelly and style it by own? You can reuse the code, but name it differently, include it in our .less files, and we are done.

Kamahl19:
+1 to Senky

PR #189, opened 2013-01-02
Hide notification brackets for guests

Kamahl19:
http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4705

### DIFF
--- a/install_instructions/install_mod.xml
+++ b/install_instructions/install_mod.xml
@@ -118,7 +118,7 @@ Please note, that if you want to use "Fetch page" feature to fetch remote pages 
 			</edit>
 			<edit>
 				<find><![CDATA[			<ul class="linklist navlinks">]]></find>
-				<action type="after-add"><![CDATA[			  <!-- IF SN_MODULE_ACTIVITYPAGE_ENABLED --><li class="icon-activitypage"><a href="{U_SN_ACTIVITYPAGE}">{L_SN_AP_ACTIVITYPAGE}</a><span id="sn-ntf-notify">(<a href="{U_VIEW_NOTIFY}">{S_SN_USER_UNREAD_NOTIFY}</a>)</span></li><!-- ENDIF -->]]></action>
+				<action type="after-add"><![CDATA[			  <!-- IF SN_MODULE_ACTIVITYPAGE_ENABLED --><li class="icon-activitypage"><a href="{U_SN_ACTIVITYPAGE}">{L_SN_AP_ACTIVITYPAGE}</a><!-- IF SN_MODULE_NOTIFY_LOADED --><span id="sn-ntf-notify">(<a href="{U_VIEW_NOTIFY}">{S_SN_USER_UNREAD_NOTIFY}</a>)</span><!-- ENDIF --></li><!-- ENDIF -->]]></action>
 			</edit>
 			<edit>
 				<find><![CDATA[				<li class="icon-ucp">]]></find>
--- a/install_instructions/update_from_0.7.2_to_0.7.3.xml
+++ b/install_instructions/update_from_0.7.2_to_0.7.3.xml
@@ -75,6 +75,12 @@ Feel free to ask for support or request more features here http://phpbbsocialnet
 				<action type="replace-with"><![CDATA[<!-- IF SN_MODULE_IM_LOADED -->]]></action>
 			</edit>
 		</open>
+		<open src="styles/prosilver/template/overall_header.html">
+      <edit>
+        <find><![CDATA[ <!-- IF SN_MODULE_ACTIVITYPAGE_ENABLED --><li class="icon-activitypage"><a href="{U_SN_ACTIVITYPAGE}">{L_SN_AP_ACTIVITYPAGE}</a><span id="sn-ntf-notify">(<a href="{U_VIEW_NOTIFY}">{S_SN_USER_UNREAD_NOTIFY}</a>)</span></li><!-- ENDIF -->]]></find>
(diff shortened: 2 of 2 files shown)


install_instructions/update_from_0.7.2_to_0.7.3.xml
```diff
@@ -0,0 +1,65 @@
...
+		</installation>
+	</header>
+	<action-group>
+		<copy>
+			<file from="root/*.*" to="*.*"/>
+		</copy>
+		<open src="styles/prosilver/template/overall_header.html">
+			<edit>
+				<find><![CDATA[			  <!-- IF SN_MODULE_ACTIVITYPAGE_ENABLED --><li class="icon-activitypage"><a href="{U_SN_ACTIVITYPAGE}">{L_SN_AP_ACTIVITYPAGE}</a><span id="sn-ntf-notify">(<a href="{U_VIEW_NOTIFY}">{S_SN_USER_UNREAD_NOTIFY}</a>)</span></li><!-- ENDIF -->]]></find>
+				<action type="replace-with"><![CDATA[			  <!-- IF SN_MODULE_ACTIVITYPAGE_ENABLED --><li class="icon-activitypage"><a href="{U_SN_ACTIVITYPAGE}">{L_SN_AP_ACTIVITYPAGE}</a><!-- IF S_SN_USER_UNREAD_NOTIFY != '' --><span id="sn-ntf-notify">(<a href="{U_VIEW_NOTIFY}">{S_SN_USER_UNREAD_NOTIFY}</a>)</span><!-- ENDIF --></li><!-- ENDIF -->]]></action>
```
senky: This is not good solution. `S_SN_USER_UNREAD_NOTIFY` always returns some data, but because of unclear data type returned in `socialnet_notify::ntf_notify_count`, there may be empty string, 0 as integer, or null. Anyway, `S_SN_USER_UNREAD_NOTIFY` is always set, and that is wrong. You should use `S_USER_LOGGED_IN` or better `SN_MODULE_NOTIFY_ENABLED`, because not all logged in users may have permissios to use notification system.

install_instructions/install_mod.xml
```diff
@@ -118,7 +118,7 @@ Please note, that if you want to use "Fetch page" feature to fetch remote pages
 			</edit>
 			<edit>
 				<find><![CDATA[			<ul class="linklist navlinks">]]></find>
-				<action type="after-add"><![CDATA[			  <!-- IF SN_MODULE_ACTIVITYPAGE_ENABLED --><li class="icon-activitypage"><a href="{U_SN_ACTIVITYPAGE}">{L_SN_AP_ACTIVITYPAGE}</a><span id="sn-ntf-notify">(<a href="{U_VIEW_NOTIFY}">{S_SN_USER_UNREAD_NOTIFY}</a>)</span></li><!-- ENDIF -->]]></action>
+				<action type="after-add"><![CDATA[			  <!-- IF SN_MODULE_ACTIVITYPAGE_ENABLED --><li class="icon-activitypage"><a href="{U_SN_ACTIVITYPAGE}">{L_SN_AP_ACTIVITYPAGE}</a><!-- ISN_MODULE_NOTIFY_ENABLED --><span id="sn-ntf-notify">(<a href="{U_VIEW_NOTIFY}">{S_SN_USER_UNREAD_NOTIFY}</a>)</span><!-- ENDIF --></li><!-- ENDIF -->]]></action>
```
senky: `SN_MODULE_NOTIFY_ENABLED`
instead of
`ISN_MODULE_NOTIFY_ENABLED`

senky: Below as well.

Please, check your code before you push it.

Culprit-CZ: I think it should be
`<!-- IF SN_MODULE_NOTIFY_ENABLED -->`

install_instructions/update_from_0.7.2_to_0.7.3.xml
```diff
@@ -70,6 +70,12 @@ Feel free to ask for support or request more features here http://phpbbsocialnet
 				<action type="replace-with"><![CDATA[<!-- IF SN_MODULE_IM_LOADED -->]]></action>
 			</edit>
 		</open>
+		<open src="styles/prosilver/template/overall_header.html">
+      <edit>
+        <find><![CDATA[        <!-- IF SN_MODULE_ACTIVITYPAGE_ENABLED --><li class="icon-activitypage"><a href="{U_SN_ACTIVITYPAGE}">{L_SN_AP_ACTIVITYPAGE}</a><span id="sn-ntf-notify">(<a href="{U_VIEW_NOTIFY}">{S_SN_USER_UNREAD_NOTIFY}</a>)</span></li><!-- ENDIF -->]]></find>
+        <action type="replace-with"><![CDATA[        <!-- IF SN_MODULE_ACTIVITYPAGE_ENABLED --><li class="icon-activitypage"><a href="{U_SN_ACTIVITYPAGE}">{L_SN_AP_ACTIVITYPAGE}</a><!-- IF SN_MODULE_NOTIFY_ENABLED --><span id="sn-ntf-notify">(<a href="{U_VIEW_NOTIFY}">{S_SN_USER_UNREAD_NOTIFY}</a>)</span><!-- ENDIF --></li><!-- ENDIF -->]]></action>
```
senky: `SN_MODULE_NOTIFY_ENABLED` must be replaced with `SN_MODULE_NOTIFY_LOADED` to get this work properly. The trick is, that module is not considered being loaded, when you have no permission to access it. But it is considered enabled. Because module is enabled, when it is enabled in the DB. But it is loaded only if it is enabled AND you have permission to access it. So use LOADED instead of ENABLED. And check your code before you push it, please.

woipi90:
hm why are my requests in this topic?!
the condition seems to be fine now..

senky:
Because @Kamahl19 accidentally included the commits with your comments to this PR.

woipi90:
so but this is still not fixed here am i right?

senky:
You are right.

woipi90:
isnt it working just fine..
what tasks still have to be changed in this pr?

senky:
This PR includes more commits and code changes that it should.

Culprit-CZ:
MA

senky:
MA

balboag:
Why this isn't fixed on dev board?

senky:
This PR includes install and update instructions only. It will need to be updated to 0.7.3 to see the changes, or manually updated. @Culprit-CZ, would you mind updating dev board to reflect this changes, please?

Culprit-CZ:
Updated.

PR #198, opened 2013-01-06
Allow php to display date of birthdays as "today".

senky:
Change `j F` to `|j F|`. With this php will output today's date as "today" instead of date itself.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4737


Culprit-CZ:
MA

Kamahl19:
MA

PR #199, opened 2013-01-08
Remove unnecessary parametres from logout link in menu block

senky:
For now, logout link in menu block looks like this:
`ucp.php?mode=logout&sid=90e43a7fc3f444857426fa8b45deef22&redirect=./activitypage.php`, but `&redirect=./activitypage.php` part is unnecessary. It should be removed.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4735

### DIFF
--- a/root/styles/prosilver/template/socialnet/block_menu.html
+++ b/root/styles/prosilver/template/socialnet/block_menu.html
@@ -49,7 +49,7 @@
 
 		<!-- IF S_USER_LOGGED_IN -->
 			<li class="ui-menu-item">
-				<a href="{U_LOGIN_LOGOUT}&amp;redirect=./activitypage.php"><span class="sn-menu-icon ui-icon ui-icon-power"></span>{L_SN_AP_LOG_OUT}</a>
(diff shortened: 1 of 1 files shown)


Kamahl19:
MA

root/styles/prosilver/template/socialnet/block_menu.html
```diff
@@ -49,7 +49,7 @@
 
 		<!-- IF S_USER_LOGGED_IN -->
 			<li class="ui-menu-item">
-				<a href="{U_LOGIN_LOGOUT}&amp;redirect=./activitypage.php"><span class="sn-menu-icon ui-icon ui-icon-power"></span>{L_SN_AP_LOG_OUT}</a>
+				<a href="{U_LOGIN_LOGOUT}"><span class="sn-menu-icon ui-icon ui-icon-power"></span>{L_SN_AP_LOG_OUT}</a>
```
senky, line 52: For you to be informed: watch condition above (I mean this: `<!-- IF S_USER_LOGGED_IN -->`). So link will for sure be always only logout link. And logout link does not accept `redirect` parametre, nor uses it to redirect user. This means that removed part is totally unnecessary. This is why it was removed, not wrapped to condition.

Culprit-CZ:
MA

PR #190, opened 2013-01-08
0 is accepted in UCP - personal info but not displayed on AP and profile

senky:
http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4701


Culprit-CZ:
MA

Kamahl19:
MA

PR #191, opened 2013-01-08
Add spaces when editing user info via profile.php

senky:
Check for [this](http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4697). It should be solved once this is merged.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4703


Culprit-CZ:
MA

Kamahl19:
MA

PR #193, opened 2013-01-08
Hover on status / comment displays the Delete button and moves the text

senky:
It should not move the text when delete button is displayed


Culprit-CZ:
MA

Kamahl19:
MA

PR #195, opened 2013-01-08
Change "Can use User status" permission to "Can write statuses"

senky:
Now if user does not have status permiossion he can not see any activity on user profile. I think this permission should only be applied on writing statuses. 

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4715


Culprit-CZ:
MA

Kamahl19:
MA

PR #203, opened 2013-01-08
It is possible to see success message when adding non-existing user

senky:
When you submit form to add new users with non-existing users filled, you gent correct error message AND success message. Also correct other messages regarding FMS.


Culprit-CZ:
MA

Kamahl19:
MA

PR #201, opened 2013-01-11
Change Activity page for guests

Kamahl19:
Hide blocks on AP.

Display only login and welcome text

http://phpbbsocialnetwork.com/viewtopic.php?f=6&t=4685

### DIFF
--- a/root/adm/style/acp_socialnet.html
+++ b/root/adm/style/acp_socialnet.html
@@ -210,13 +210,35 @@
 <!-- ENDIF -->
 
 <!-- IF ACP_SN_MANAGE eq 'welcome' -->
-	{L_ACP_SN_ACTIVITYPAGE_WELCOME_EXPLAIN}<br /><br />
-
-	{L_ACP_SN_ACTIVITYPAGE_WELCOME_INSTRUCTIONS}
-
-	<fieldset><code>'SN_AP_WELCOME_TITLE'</code></fieldset>
-
-	{L_ACP_SN_ACTIVITYPAGE_WELCOME_EDIT}
+	<form id="postform" name="postform" method="post" action="{U_ACTION}&amp;manage=welcome">
+		<fieldset>
+		  <legend>{L_ACP_SN_WELCOME_TEXT}</legend>
+		  <dl>
+				<dt><label for="welcome_text_title">{L_ACP_SN_WELCOME_TEXT_TITLE}:</label></dt>
+				<dd><input type="text" id="welcome_text_title" name="welcome_text_title" size="40" value="{WELCOME_TEXT_TITLE}"/></dd>
+			</dl>
+			<dl>
+				<!-- INCLUDE ./acp_socialnet_activitypage_bbcode.html -->
+				<textarea name="welcome_text" id="welcome_text" rows="15" cols="70" style="width:95%;font-size: 11px;margin-top: 5px;">{WELCOME_TEXT}</textarea>
+			</dl>
+			<!-- IF WELCOME_TEXT_PREVIEW -->
+				<table cellspacing="1">
+					<thead>
+						<tr><th colspan="2">{L_PREVIEW}</th></tr>
+					</thead>
+					<tbody>
+						<tr>
+							<td>{WELCOME_TEXT_PREVIEW}</td>
+						</tr>
+					</tbody>
+				</table>
+			<!-- ENDIF -->
+			<p class="submit-buttons" style="font-size: 12px;">
+				<input class="button1" type="submit" id="submit" name="submit" value="{L_SUBMIT}" />&nbsp;
+				<input class="button2" type="reset" id="reset" name="reset" value="{L_RESET}" />
+			</p>
+		</fieldset>
+	</form>
 <!-- ENDIF -->
 
 <!-- IF B_ACP_SN_PROFILE -->
--- a/root/adm/style/acp_socialnet.html
+++ b/root/adm/style/acp_socialnet.html
@@ -210,7 +210,7 @@
 <!-- ENDIF -->
 
 <!-- IF ACP_SN_MANAGE eq 'welcome' -->
-	<form id="postform" name="postform" method="post" action="{U_ACTION}&amp;manage=welcome">
+	<form id="postform" name="postform" method="post" action="{U_ACTION_WELCOME}">
 		<fieldset>
 		  <legend>{L_ACP_SN_WELCOME_TEXT}</legend>
 		  <dl>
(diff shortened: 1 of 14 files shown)


root/adm/style/acp_socialnet.html
```diff
@@ -210,13 +210,35 @@
...
 
 <!-- IF ACP_SN_MANAGE eq 'welcome' -->
-	{L_ACP_SN_ACTIVITYPAGE_WELCOME_EXPLAIN}<br /><br />
-
-	{L_ACP_SN_ACTIVITYPAGE_WELCOME_INSTRUCTIONS}
-
-	<fieldset><code>'SN_AP_WELCOME_TITLE'</code></fieldset>
-
-	{L_ACP_SN_ACTIVITYPAGE_WELCOME_EDIT}
+	<form id="postform" name="postform" method="post" action="{U_ACTION}&amp;manage=welcome">
```
senky: Add a proper template variable for `action`. Do not hard-code it.

root/adm/style/acp_socialnet_activitypage_bbcode.html
```diff
@@ -0,0 +1,103 @@
+<script type="text/javascript">
```
senky, line 2: I am not sure, but isn't this file exact copy of bbcode.html in template of prosilver? Why do not you use that file instead of creating new copy of it?

root/language/en/mods/socialnet.php
```diff
@@ -79,6 +73,7 @@
 	'SN_CB_FETCH_ERROR'						 => 'An error was encountered when fetching the web page',
 
 	'SN_AP_ACTIVITYPAGE'					 => 'Activity page',
+	'SN_AP_CREATE_ACCOUNT'					=> 'Create Account',
```
senky, line 76: Huh?

root/socialnet/acp/acp_activitypage.php
```diff
@@ -80,6 +79,60 @@ function main($id)
...
+      $user->add_lang('posting');
+
+			include($phpbb_root_path . 'includes/functions_display.' . $phpEx);
+  		include($phpbb_root_path . 'includes/message_parser.' . $phpEx);
+
+			$submit	= (isset($_POST['submit'])) ? true : false;
+
+      if ($submit)
+			{
+			  $welcome_text_title = request_var('welcome_text_title', '');
```
senky: This should normalize utf8 as well.

senky:
I am really not sure welcome text requires additional DB table. Just include all colums to sn config table with proper naming and that is it.

root/styles/prosilver/template/socialnet/activitypage_body.html
```diff
@@ -3,100 +3,96 @@
...
+			<!-- ELSEIF S_SN_AP_NOTIFY -->
 
 				<!-- INCLUDE socialnet/activitypage_notify.html -->
 
-		<!-- ELSEIF not S_USER_LOGGED_IN -->
-
-				<!-- INCLUDE socialnet/block_login.html -->
+			<!-- ENDIF -->
+		</div>
+	<!-- ENDIF -->
```
senky: Just replace this line with `<!-- ELSE -->` and delete the line below. Please, combine conditions if possible.

root/adm/style/acp_socialnet_activitypage_bbcode.html
Kamahl19, line 2: What if there is no prosilver based style? Or what if some custom style has different name of textarea.. We need to copy the prosilver original code so we are sure it will always work

root/language/en/mods/socialnet.php
Kamahl19, line 76: Whats wrong with that. Do you want to use Register? I think this is much better, people are used to it.

Kamahl19:
We can not save it to config table because there could be a long text and we use varchar(255) for config table.

Maybe we could create another config table for text values so we can use it for something else in future too.

root/adm/style/acp_socialnet_activitypage_bbcode.html
senky, line 2: If it has different name of textarea, it will not work for whole phpBB. So please do not construct stupid cases. Please, reuse what can be reused. Do not create duplicate code. It is enough of that in the SN code without adding new one.

root/language/en/mods/socialnet.php
senky, line 76: Why do we need to have everything different from the native phpBB? What about trying to reuse as much of phpBB as we can? Including language keys.

senky:
Hm, that is really bad. But yes, mabye creating new config table, or replace current config table colum with TEXT type is much better idea, than includig special table for welcome text only... I think modifying current table would not be as much harmy, because it is selected only once a script, plus cached.

root/language/en/mods/socialnet.php
Kamahl19, line 76: Why do you think phpbb does something good way? We are here to improve it, customize it. People are used to "Create Account", Register was popular 6 years ago when phpBB3 was released. Do you really want to reuse as much as possible from 6-7 years old project?

root/adm/style/acp_socialnet_activitypage_bbcode.html
Kamahl19, line 2: I dont think we should connect style with ACP style. These are 2 different parts of the system. You can use not prosilver based style or you can use heavily modified prosilver based style and it will not work. If there is different name for form_name = 'postform'; it wont work in ACP but it will work in style. And yes I have seen prosilver based styles which changed this "postform" in JS and in html too. What then ?

root/adm/style/acp_socialnet.html
```diff
@@ -210,13 +210,35 @@
...
-	{L_ACP_SN_ACTIVITYPAGE_WELCOME_EDIT}
+	<form id="postform" name="postform" method="post" action="{U_ACTION_WELCOME}">
+		<fieldset>
+		  <legend>{L_ACP_SN_WELCOME_TEXT}</legend>
+		  <dl>
+				<dt><label for="welcome_text_title">{L_ACP_SN_WELCOME_TEXT_TITLE}:</label></dt>
+				<dd><input type="text" id="welcome_text_title" name="welcome_text_title" size="40" value="{WELCOME_TEXT_TITLE}"/></dd>
+			</dl>
+			<dl>
+				<!-- INCLUDE ./acp_socialnet_activitypage_bbcode.html -->
```
gooof, line 221: use only `<!-- INCLUDE acp_socialnet_activitypage_bbcode.html -->`

root/adm/style/acp_socialnet_activitypage_bbcode.html
gooof, line 2: Please use code parts from the `acp_users_signature.html` file in adm/style.

root/language/en/mods/socialnet.php
senky, line 76: Ok then.

PR #204, opened 2013-01-23
Update jquery to 1.9.1

Culprit-CZ:
Many deprecated features


Culprit-CZ:
Now is necessary to update all deprecated features. See: [jQuery Core 1.9 Upgrade Guide](http://jquery.com/upgrade-guide/1.9/)

We have two choices:
1) Use [jQuery migrate 1.0.0 plugin](https://github.com/jquery/jquery-migrate/)
2) Rewrite all deprecated features.

`jQuery.browser` is also deprecated, so we need find some effective solution.

Culprit-CZ:
**Still:** `JQMIGRATE: Can't change the 'type' of an input or button in IE 6/7/8`

## 

**jQuery UI** updated to 1.10

`jQuery.browser` moved to `socialNetwork.browser` - used original jQuery browser detection
**File:**
`jquery.detectmobilebrowser.min.js` rewritten as function `_detectMobileDevice()` and called from `socialNetwork`
`jquery.timers.js` removed line with `.browser`

**Update:**
`jquery.events.inpus.js`, `jquery.forms.js`, `jquery.jgrowl.min.js`

Culprit-CZ:
It seems to be everything.
`JQMIGRATE: Can't change the 'type' of an input or button in IE 6/7/8` is remaining. But it is working normal on old browsers.
Widget `.dialog()` from jQuery UI is reporting this.

**jQuery UI should be upgraded from 1.9.2 to 1.10**
Should be merged with #207

Kamahl19:
Are you sure about renaming to jquery.cookie.js ? I remember some users have had problem with this name so we renamed it to jquery_cookie.js

Culprit-CZ:
I remember it. But we should use same names for each jQuery plugin.

Kamahl19:
I know but it is causing problems on some servers

Culprit-CZ:
Than should we change `jquery_` and `jquery.` to `jquery-`
Then we can remove _versions_ from filenames, in each file exist the information about _version_.
`jquery-1.9.0.min.js` to `jquery-min.js`
`jquery-ui-1.10.0.custom.min.js`` to`jquery-ui.custom.min.js```

senky:
What problem on some servers? I cannot think of any. The only problem may be, that they use jquery cookie plugin as well. So they need to remove one. Can you, please, reffer to exact explanation of problem? We should follow the naming convenions.

Kamahl19:
no they did not have any other jquery cookie included. we did not find the problem but renaming the file solved it. This happened to several users and I remember we found out it happened only on some specific server configuration. I dont remember more. It was a long time ago and we did not used github or proper bug tracker so..

senky:
So we should name it `jquery.cookie.js` and find the core of the problem. Or minify it, name it `jquery.cookie.min.js` and hope the problem will not persist.

senky:
I have done a research. This is the topic:
http://phpbbsocialnetwork.com/viewtopic.php?f=21&t=87

Now, the problem is in `mod_security` on Apache servers. Or to be more specific, it is problem in old, not updated `mod_security`. Read:
http://drupal.org/node/522646
Drupal has reported the problem to mod_security bug tracker, and it seems to be resolved now (according to issue status). The workaround is proposed here: http://drupal.org/node/522646#comment-4327268

So I think we should rename it to `jquery.cookie.js` and rather prepare a tutorial for users with outdated mod_security, on how to use .htaccess to overcome this problem. Your opinions?

Kamahl19:
agree good job

root/socialnet/js/jquery.timers.min.js
```diff
@@ -0,0 +1,16 @@
+/*
+ Timers
```
senky: If [this](http://jquery.offput.ca/every/) is the plugin we use, we should include at least link to it.

root/socialnet/js/jquery.timers.min.js
```diff
@@ -0,0 +1,16 @@
...
+label,fn);this.inProgress=false};handler.$timerID=fn.$timerID;if(!element.$timers[label][fn.$timerID])element.$timers[label][fn.$timerID]=window.setInterval(handler,interval);if(!this.global[label])this.global[label]=[];this.global[label].push(element)},remove:function(element,label,fn){var timers=element.$timers,ret;if(timers){if(!label)for(label in timers)this.remove(element,label,fn);else if(timers[label]){if(fn){if(fn.$timerID){window.clearInterval(timers[label][fn.$timerID]);delete timers[label][fn.$timerID]}}else for(var fn in timers[label]){window.clearInterval(timers[label][fn]);
+delete timers[label][fn]}for(ret in timers[label])break;if(!ret){ret=null;delete timers[label]}}for(ret in timers)break;if(!ret)element.$timers=null}}}});
+/**
+ * @preserve Remove during jQuery update
+ * @autor Culprit
+ */
+jQuery(window).one("unload",function(){var global=jQuery.timer.global;for(var label in global){var els=global[label],i=els.length;while(--i)jQuery.timer.remove(els[i],label)}});
```
senky: If you wrote this, and it is not in deminified version elsewhere, it should not be minified, IMO.

root/socialnet/js/jquery.titlealert.min.js
```diff
@@ -0,0 +1,21 @@
...
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in
+ all copies or substantial portions of the Software.
+*/
+(function($){$.titleAlert=function(text,settings){if($.titleAlert._running)$.titleAlert.stop();$.titleAlert._settings=settings=$.extend({},$.titleAlert.defaults,settings);if(settings.requireBlur&&($.titleAlert.hasFocus||$.titleAlert.activeElement))return;settings.originalTitleInterval=settings.originalTitleInterval||settings.interval;$.titleAlert._running=true;$.titleAlert._initialText=document.title;document.title=text;var showingAlertTitle=true;var switchTitle=function(){if(!$.titleAlert._running)return;
```
senky: What is the service you use to minify JS? I suggest using [Closure Compiler](http://closure-compiler.appspot.com/home) - it can minify it even better. And minify [this](https://github.com/heyman/jquery-titlealert/blob/master/jquery.titlealert.js) file, please.

root/socialnet/js/jquery.cookie.min.js
```diff
@@ -0,0 +1,16 @@
...
+ @cat Plugins/Cookie
+ @author Klaus Hartl/klaus.hartl@stilbuero.de
+
+ Copyright (c) 2006 Klaus Hartl (stilbuero.de)
+ Dual licensed under the MIT and GPL licenses:
+ http://www.opensource.org/licenses/mit-license.php
+ http://www.gnu.org/licenses/gpl.html
+
+*/
+jQuery.cookie=function(d,c,a){if("undefined"!=typeof c){a=a||{};null===c&&(c="",a.expires=-1);var b="";if(a.expires&&("number"==typeof a.expires||a.expires.toUTCString))"number"==typeof a.expires?(b=new Date,b.setTime(b.getTime()+864E5*a.expires)):b=a.expires,b="; expires="+b.toUTCString();var e=a.path?"; path="+a.path:"",f=a.domain?"; domain="+a.domain:"";a=a.secure?"; secure":"";document.cookie=[d,"=",encodeURIComponent(c),b,e,f,a].join("")}else{c=null;if(document.cookie&&""!=document.cookie){a=
```
senky: Minify [this](https://github.com/carhartl/jquery-cookie/blob/master/jquery.cookie.js) file using [Closure Compiler](http://closure-compiler.appspot.com/home), please.

root/socialnet/js/m.socialnet.js
```diff
@@ -409,6 +411,44 @@ var socialNetwork = (function($) {
...
+
+			// Chrome is Webkit, but Webkit is also Safari.
+			if (browser.chrome) {
+				browser.webkit = true;
+			} else if (browser.webkit) {
+				browser.safari = true;
+			}
+
+			this.browser = browser;
+			this.browser.mobile = _detectMobileDevice(navigator.userAgent||navigator.vendor||window.opera);
```
senky, line 448: This line and line below can be compressed:

``` js
this.mobileBrowser = this.browser.mobile = _detectMobileDevice(navigator.userAgent||navigator.vendor||window.opera);
```

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
```diff
@@ -26,50 +26,51 @@
...
 jQuery(document).ready(function($){
 	socialNetwork.init({
 		allow_load: {B_AJAX_LOAD_ALLOW},
 		rtl: '{S_CONTENT_DIRECTION}' == 'rtl' ? true: false,
 	    expanderTextMore: '{L_SN_EXPANDER_READ_MORE}',
 	    expanderTextLess: '{L_SN_EXPANDER_READ_LESS}',
-	    browserOutdatedTitle : '{L_BROWSER_OUTDATED_TITLE}',
-	    browserOutdated : '{L_BROWSER_OUTDATED}',
+	    browserOutdatedTitle: '{L_BROWSER_OUTDATED_TITLE}',
+		browserOutdated: '{L_BROWSER_OUTDATED}',
```
senky, line 43: Replace spaces with tab, please.

root/socialnet/js/jquery.timers.min.js
Culprit-CZ: I'm really sorry. I did not find any successful link for this plugin.
You find it.

root/socialnet/js/jquery.timers.min.js
Culprit-CZ: I want only to leave these, because I have modify it. Otherwise the plugin is 2-3 years no more updated.

root/socialnet/js/jquery.titlealert.min.js
Culprit-CZ: Nothing against. I leave only comment header.
Ok, no problem to remove.

I'm using Closure compiler. But I give there for comments `@preserve`

root/socialnet/js/jquery.cookie.min.js
Culprit-CZ: Closure compiler is used.

root/socialnet/js/jquery.timers.min.js
senky: Use this:
http://jquery.offput.ca/every/
or this:
http://archive.plugins.jquery.com/project/timers

root/socialnet/js/jquery.form.js
```diff
@@ -1,467 +1,603 @@
 /*!
```
senky, line 1: This plugin can be minified as well, I believe.

root/socialnet/js/jquery.timers.min.js
```diff
@@ -0,0 +1,12 @@
...
+/*! Remove during jQuery update */
+jQuery(window).one("unload", function() {
```
senky: Please, look at original, non-minified code. It `uses jQuery.browser.msie`. It needs to be replaced by `$sn.browser`.

And remove lines 2-12, they are just duplicates of the last stuff done in the first line. Whatsmore, this one does not contain condition to be ran only in MSIE.

Culprit-CZ:
WE could not use `$sn.browser` by plugin load. `$sn` is not defined.

`if (/msie/.exec(navigator.userAgent.toLowerCase() || false) !== false )`
`if(!1!==(/msie/.exec(navigator.userAgent.toLowerCase())||!1))`

woipi90:
just wanted to inform u that 1.9.1 is released..

maybe u can even update to this version..

root/socialnet/js/jquery.timers.min.js
```diff
@@ -0,0 +1,2 @@
...
+if(!1!==(/msie/.exec(navigator.userAgent.toLowerCase())||!1))jQuery(window).one("unload",function(){var b=jQuery.timer.global,a;for(a in b)for(var c=b[a],d=c.length;--d;)jQuery.timer.remove(c[d],a)});
```
senky: Why is that `||!1` at the end of the condition?

senky:
Yes, it is good idea to update it directly to 1.9.1.

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
```diff
@@ -1,19 +1,19 @@
...
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.form.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.titlealert.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.detectmobilebrowser.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.form.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.titlealert.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.expander.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.editable.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.jgrowl_minimized.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.jgrowl.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/underscore-min.js"></script>
```
woipi90, line 23: maybe should be also renamed to *.min.js instead of *-min.js to have an unified ending..

senky, line 23: Good point! @Culprit-CZ, please, manage it, and do not forget about https://github.com/phpBB-Social-Network/phpBB-Social-Network/pull/204#discussion_r2908902

woipi90:
if there are allready nearly all jquery files updated in this pr maybe u can add this one for fixing the flickering of mentions inputbox..

https://github.com/podio/jquery-mentions-input/commit/5df262507feb00ac2e732203e188bf033e0ebbe0

root/socialnet/js/jquery.timers.min.js
Culprit-CZ: `/msie/.exec(navigator.userAgent.toLowerCase()` returns `null` when string `msie` is not found
`(/msie/.exec(navigator.userAgent.toLowerCase())||!1)` return `true` when `msie` is found in `navigator.userAgent`, return `false` when string `msie` is not found

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
```diff
@@ -1,20 +1,20 @@
...
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.detectmobilebrowser.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.form.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.titlealert.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.expander.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.editable.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.jgrowl_minimized.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/underscore-min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.jgrowl.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/underscore.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.mentionsInput.js"></script>
```
senky, line 24: jquery.mentionsInput.js can be minified as well.

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
```diff
@@ -1,20 +1,20 @@
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery-1.8.3.min.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.metadata.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery_cookie.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.timers.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.detectmobilebrowser.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.metadata.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.cookie.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.timers.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.hoverIntent.min.js"></script>
```
woipi90, line 12: i dont know exactly but if thats the original file then there would be an update for this..
here: http://cherne.net/brian/resources/jquery.hoverIntent.minified.js

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
```diff
@@ -1,20 +1,20 @@
...
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.text-overflow.min.js"></script>
```
woipi90, line 13: there are also a view small changes here and 4v updates..
here: https://github.com/bramstein/text-overflow/blob/master/lib/jquery.text-overflow.js

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
```diff
@@ -1,20 +1,20 @@
...
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.watermark.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery-ui-1.9.2.custom.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.elastic.source.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.events.input.js"></script>
```
woipi90, line 17: some browsercompatibility fixes (f. opera)..
here: https://github.com/dodo/jquery-inputevent/blob/master/jquery.inputevent.js

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
Culprit-CZ, line 12: We do not use this. We use another hoverIntent http://pupunzi.com

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
Culprit-CZ, line 17: Will be upgraded trough #207

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
woipi90, line 12: ok so its a modified version by this author.. i ddnt exactly know if its only used in the mods or if its modified itself..

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
woipi90, line 17: ah ok and i guess it will be minified there.. 
other two files not min are editable and elastic.source..

senky, line 17: Elastic is edited to fit SN, so it cannot be minified unless there will be a developer's version packed, too (which is, in fact, not a bad idea!).
Editable is entirely developed by us, so that is why it is not minified (but idea to provide users with minified version and include .dev.js versions for firther developments is really not bad idea!)

senky:
MA

PR #207, opened 2013-01-30
Update jQuery UI 1.10

Culprit-CZ:
In new version are many things deprecated.
We need to upgrade to 1.10 when we upgraded jQuery to 1.9.0 with #204 
Here is [jQuery UI 1.10 upgrade guide](http://jqueryui.com/upgrade-guide/1.10/)
Should be merged with #204 
After jQuery upgrade we become this migration message `JQMIGRATE: jQuery.fn.andSelf() replaced by jQuery.fn.addBack()`


Kamahl19:
MA

root/socialnet/js/m.profile.js
```diff
@@ -53,30 +53,17 @@
...
 					if (event.which == 3)
 						return false;
-				}).tabs({
-					ajaxOptions: {
-						cache: false,
-						async: false,
-						data: {
-							fullPage: self.tabShowFullPage
-						}
+				}).removeAttr('style').tabs({
```
senky: I am not sure if removing style before triggering tabs is a good step. There will be a while, when elements will be not styled at all.

root/socialnet/js/m.profile.js
```diff
@@ -53,30 +53,17 @@
...
+					beforeLoad: function(e, ui) {
+						ui.ajaxSettings.cache = false;
+						ui.ajaxSettings.async = false;
+						ui.ajaxSettings.url += '&fullPage=' + self.tabShowFullPage;
 					},
-					cache: false,
```
senky, line 64: You removed it, but I cannot see it redefined elsewhere. Tabs should not be cached, so please include it in ajaxSettigs.

root/socialnet/js/m.profile.js
```diff
@@ -126,7 +113,7 @@
 
 					}
 				})
-				.removeAttr('style');
+
```
senky: Remove this line, please. It is unnecessary doubleline here.

root/socialnet/js/m.profile.js
```diff
@@ -53,30 +53,17 @@
...
-					create: function(e, ui) {
-						if ($sn.getCookie('sn-up-profileTab') == 0) {
-							return false;
-						}
-					},
-					select: function(e, ui) {
-						$(ui.panel).html(self.spinner);
+					beforeActivate: function(e, ui) {
+						ui.newPanel.html(self.spinner);
+						$sn.setCookie('sn-up-profileTab',ui.newTab.index());
```
senky: Space after comma, please:

``` js
$sn.setCookie('sn-up-profileTab', ui.newTab.index());
```

root/socialnet/js/m.profile.js
Culprit-CZ, line 64: Redefined by using `$sn.getCookie` & `$sn.setCookie`

senky:
We can add jQuery UI css to stylesheet.css, so that it gets automatically merged to all other styles and does not consume one extra request for browser.

Culprit-CZ:
We could not. Than we have more work with upgrading.
jQuery UI CSS use: `images/some_image.png`
phpBB CSS use: `{T_THEME_PATH}/images/some_image.png`

If we do, then we must have in mind the editing CSS.
Thus, we allow a simple change of style with [jQuery UI ThemeRoller](http://jqueryui.com/themeroller/)

Kamahl19:
Looks good, MA

senky:
MA

PR #202, opened 2013-02-01
Mark all notifications from status you opened 

Culprit-CZ:
When you open the status from notification, it should mark all your notifications from this status. 

Now, when you have 2 new comments on your status and you open one ntf and read both comments, it does not mark the other ntfs

### DIFF
--- a/root/socialnet/notify.php
+++ b/root/socialnet/notify.php
@@ -7,7 +7,6 @@
  * @license http://opensource.org/licenses/gpl-license.php GNU Public License
  *
  */
-
 if (!defined('SOCIALNET_INSTALLED'))
 {
 	/**
@@ -30,7 +29,6 @@ if (!defined('SOCIALNET_INSTALLED'))
 	$user->session_begin(false);
 	$auth->acl($user->data);
 	$user->setup();
-
 }
 
 if (!class_exists('socialnet_notify'))
@@ -54,12 +52,12 @@ if (!class_exists('socialnet_notify'))
 		 */
 		function socialnet_notify(&$p_master)
 		{
-			$this->p_master =& $p_master;
+			$this->p_master = & $p_master;
 		}
 
 		function init()
 		{
-			global $user, $db, $phpbb_root_path, $phpEx, $template, $config;
+			global $phpbb_root_path, $phpEx, $template, $config;
 			$this->time = time();
 
 			$this->ntf_delete();
@@ -71,11 +69,11 @@ if (!class_exists('socialnet_notify'))
 			$this->ntf_ap_show();
 
 			$template->assign_vars(array(
-				'U_VIEW_NOTIFY'				 			=> append_sid("{$phpbb_root_path}activitypage.$phpEx", 'mode=notify'),
-				'S_SN_USER_UNREAD_NOTIFY'		=> $this->ntf_notify_count(),
-				'S_SN_NTF_THEME'			 			=> $config['ntf_theme'],
-				'SN_NTF_LIFE'				 				=> $config['ntf_life'] * 1000,
-				'SN_NTF_CHECKTIME'			 		=> $config['ntf_checktime'] * 1000,
+				'U_VIEW_NOTIFY'				 => append_sid("{$phpbb_root_path}activitypage.$phpEx", 'mode=notify'),
+				'S_SN_USER_UNREAD_NOTIFY'	 => $this->ntf_notify_count(),
+				'S_SN_NTF_THEME'			 => $config['ntf_theme'],
+				'SN_NTF_LIFE'				 => $config['ntf_life'] * 1000,
+				'SN_NTF_CHECKTIME'			 => $config['ntf_checktime'] * 1000,
 			));
 		}
 
@@ -97,7 +95,6 @@ if (!class_exists('socialnet_notify'))
 			/**
 			 * @ignore
 			 */
-
 			$ntf_type = request_var('type', 'check');
 			$ntf_id = request_var('nid', 0);
 			if ($ntf_type == 'delete' && $ntf_id != 0)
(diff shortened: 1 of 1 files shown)


root/socialnet/notify.php
```diff
@@ -364,10 +365,12 @@ function ntf_check_MARK()
 				return;
 			}
 
+			// Mark readed all other notifications to same status, etc.
+			$ntf_like = str_replace('&', '&amp;', preg_replace('/ntfMark=[0-9]+&/','', $_SERVER['QUERY_STRING']));
```
senky: Using `$_SERVER` is unsecure. But what is even more unsecure is that you do not escape the string before you perform your SQL query. See https://github.com/phpBB-Social-Network/phpBB-Social-Network/pull/202/files#L0R373. So please, stop using `$_SERVER` (because it can be easily replaced by custom string) and excape all inputs to SQL queries.

gooof:
Can you also checkout the 4 tasks here: http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4742

root/socialnet/notify.php
gooof: What shows `$_SERVER['QUERY_STRING']` here?

senky:
> Can you also checkout the 4 tasks here: http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4742

This will be solved using new PR. One PR = one issue.

root/socialnet/notify.php
Culprit-CZ: I removed `$db->sql_escape`, you have right.
Show me, how to get whole parameters, that are posted

root/socialnet/notify.php
```diff
@@ -364,10 +365,20 @@ function ntf_check_MARK()
 				return;
 			}
 
+			// Mark readed all other notifications to same status, etc.
+			unset($_GET['ntfMark']);
+			$params = array();
+			
+			foreach( $_GET as $id => $val)
```
senky: I believe all parametres posted using GET can be predictable. So you can `request_var()` them, so that it is secured.

Culprit-CZ: Could not be predictable.
At this moment could be predictable, but when you expand notifications for some another module, than you become some another parameters.

senky: So what about adding new public variable to `socialnet_notify` class and new public method to add new parametres on the fly? Merging hooks, it will be even easier, but it is possible even now. It will be just much much more obvious and expansible, than this.

senky:
MA

gooof:
no escape?
and check if its not always mark as read.

gooof:
What is when a link is empty?

Check if status_id and ntf_mark is set, if true run the update query. Therefore you need only a new db column, but it pays.

-- edit
Or a new way. Without anything, you need only check `ntf_data` == `ntf_data` ;)

Culprit-CZ:
By notification could not be empty link. Because the link give the page where should be redirected.

gooof:
Function `emote()` has an empty link. 

Also `ntfMark=0` = empty link (not 0 i mean if ID not exists)

And there is no sql_freeresult

Culprit-CZ:
emote has a empty link because you can not view somewhere else than in notify page.

senky:
But that is the point! Script will try to mark all notifications with the same link as read. But as there will be always true condition `WHERE ntf_data LIKE '%%'`, it will mark **all** notifications as read. So there should be a condition that will allow query to be ran only if link is not empty string.

gooof:
In my small notify table do this sql query the right job:

``` mysql
SELECT n.ntf_data, n.ntf_user, n.ntf_read, a.ntf_data
FROM `tippsundtricks_sn_notify` AS n,
   (SELECT ntf_data FROM `tippsundtricks_sn_notify` WHERE ntf_id = 12 AND ntf_user = 71) AS a
WHERE n.ntf_user = 71
    AND n.ntf_read != 'SN_NTF_STATUS_READ'
    AND n.ntf_data = a.ntf_data
```

the same as update:

``` mysql
UPDATE `tippsundtricks_sn_notify` AS n,
   (SELECT ntf_data FROM `tippsundtricks_sn_notify` WHERE ntf_id = 12 AND ntf_user = 71) AS a
SET n.ntf_user = 2
WHERE n.ntf_user = 71
    AND n.ntf_read != 'SN_NTF_STATUS_READ'
    AND n.ntf_data = a.ntf_data
```

inside function use only UPDATE, and don't forget to change the table and values.
Much faster than SELECT and LIKE.

senky:
But the problem is, that you cannot use exact comparing for ntf_data. The only part they share is ntf_link. So exact comparison would not work. Am I right @Culprit-CZ? Or are they exactly same? ...hm.

gooof:
hm yes you are right, thats not good... usernames are different.
Then we really need a db column for `status_id` to make more than one notify as read at the same time.
LIKE queries are too slow on large tables.

Culprit-CZ:
@Senky Are not exactly same, users could be different
@gooof We should not add any more columns to DB table. This DB table should be used for any another module which will use notifications.

senky:
So please add empty check and run the query only if ntf_url is not empty.

root/socialnet/notify.php
```diff
@@ -364,10 +365,19 @@ function ntf_check_MARK()
 				return;
 			}
 
+			$sql = "SELECT ntf_data FROM " . SN_NOTIFY_TABLE . " WHERE ntf_id = '{$ntf_mark}'";
+			$rs = $db->sql_query( $sql);
+			$data = unserialize($db->sql_fetchfield('ntf_data', $rs));
```
senky: Now we need to add some more logic. Verify, if at least one row has been affected, instead of `!isset($data['link'])` check (in you do not do that, it is possible, that php will throw an error):

``` php
if ( $db->sql_affectedrows() ) 
```

Inside the condition body, add new condition (and please, use `!= ''` instead of `empty()` check:

``` php
if ( $data['link'] != '' )
```

and now, inside the second condition, you can run the query.

senky:
MA

Kamahl19:
MA

PR #209, opened 2013-02-04
Patch 3

woipi90:


Kamahl19:
what is this?

woipi90:
sry i think i actually messed up again.. only wanted to add my commit to pr 207 and somehow it split..

senky:
You cannot add commits to any PR. So either comment on the PR and an author will fix it, or send a new PR, which needs to be properly named, described, labeled, milestoned and each commit in it needs to start with `[i/<PR ID>]`.

woipi90:
ok will only comment in future.. ;) sry about this..

so what should i do about this?

Culprit-CZ did include most of my changes anyway in his latest changes in overall_header_hook_js..

senky:
I will close this. Really thanks for your interest. I am looking forward your comments on opened PRs.

PR #210, opened 2013-02-05
Remove mainpage.php

Kamahl19:
Do not forget to include deletion to /install_instructions/update_*.xml file!


senky:
I can see there are some JS files removed, even thought it has nothing to do with mainpage.php (so my opinion is not to include this change to this PR), but there is no mainpage.php removal!

Culprit-CZ:
MA

senky:
MA

PR #206, opened 2013-02-05
Add link to change log to ACP update message

Kamahl19:
We should add a link to changelog to ACP when text about new version is displayed. It could help persuade admins to update

http://phpbbsocialnetwork.com/viewtopic.php?f=6&t=4799

### DIFF
--- a/root/adm/style/socialnet_update.html
+++ b/root/adm/style/socialnet_update.html
@@ -5,7 +5,7 @@
 		<div class="errorbox">
 			<p>{L_ACP_SN_VERSION_COULD_NOT_GET_INFORMATION}</p>
 		</div>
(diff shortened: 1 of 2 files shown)


root/adm/style/socialnet_update.html
```diff
@@ -2,10 +2,10 @@
 	<p>{L_SN_VERSION_CHECK_EXPLAIN}</p>
 
 	<!-- IF B_COULD_NOT_GET_UPDATE -->
-		<div class="errorbox">
+  	<div class="errorbox">
```
senky: If you want to replace it, the only allowed indention character is tab, not spaces.

Culprit-CZ:
MA

senky:
MA

PR #212, opened 2013-02-05
Add Twitter Follow button to ACP

Kamahl19:
http://phpbbsocialnetwork.com/viewtopic.php?f=78&t=4801


senky:
Are you sure we want so big button?
For me, it would be much better to have it as small as Facebook like button. This one is too big even for blind one! Medium (default) size seems to be fine...

root/adm/style/acp_socialnet.html
```diff
@@ -264,11 +264,13 @@
...
+		<legend>{L_ACP_SN_FB_TWITTER}</legend>
+		{L_ACP_SN_FB_TWITTER_EXPLAIN}
+		<object type="text/html" data="//www.facebook.com/plugins/like.php?href=http%3A%2F%2Fwww.facebook.com%2Fpages%2FphpBB-Social-Network%2F180271885389370&amp;send=false&amp;layout=standard&amp;width=650&amp;show_faces=true&amp;action=like&amp;colorscheme=light&amp;font&amp;height=25&amp;appId=217550038308424" width="100%" height="25" style="margin: 15px 0 10px 0;">
 			<param name="src" value="//www.facebook.com/plugins/like.php?href=http%3A%2F%2Fwww.facebook.com%2Fpages%2FphpBB-Social-Network%2F180271885389370&amp;send=false&amp;layout=standard&amp;width=650&amp;show_faces=true&amp;action=like&amp;colorscheme=light&amp;font&amp;height=25&amp;appId=217550038308424" />
 		</object>
+    <a href="https://twitter.com/phpBBSocialNet" class="twitter-follow-button" data-show-count="false" data-size="large">Follow @phpBBSocialNet</a>
```
Culprit-CZ: Error: there is no attribute "data-show-count"
Error: there is no attribute "data-size"

root/adm/style/acp_socialnet.html
```diff
@@ -264,11 +264,13 @@
...
+    <script>!function(d,s,id){var js,fjs=d.getElementsByTagName(s)[0];if(!d.getElementById(id)){js=d.createElement(s);js.id=id;js.src="//platform.twitter.com/widgets.js";fjs.parentNode.insertBefore(js,fjs);}}(document,"script","twitter-wjs");</script>
```
Culprit-CZ: Error: required attribute "type" not specified

root/adm/style/acp_socialnet.html
senky: Where you can see such error? data attributes are part of HTML5. I know, that phpBB uses XHTML, but it is rather stupid to change header of phpBB just because of this errors. All browsers can manage that, not speaking that it is used in JS only. Whatsmore, Twitter does not provide an alternative solution (except for iframe, which is depreciated by W3C).

Culprit-CZ: Here can you view the problem http://users.skynet.be/mgueury/mozilla/
We could not use the HMTL5 because we still under phpBB HTML page declaration

senky: I know it throws errors, but it works. But if you want to, @Kamahl19, you can replace it with iframe.

Kamahl19: I agree with Senky, it is not a problem and there is no other solution except iframe which is really bad

root/adm/style/acp_socialnet.html
Kamahl19: same as above

Kamahl19:
I like its size. Culprit please decide whether it is better big or small so we can finish this

root/adm/style/acp_socialnet.html
senky: iframe is not bad. It keeps our code clean and has absolutelly the same functionality as html5 version. And it is valid. So maybe it should be replaced with iframe.

Kamahl19: I dont understand now. You also said iframe is deprecated so why use it when we can use modern html5? One solution is deprecated and old, the other does not fit with our doctype but is modern.

senky: iframe is depreciated in html5. However, phpBB still uses XHTML doctype (it will use html5 in 3.1), where iframe is common tag. So we should use iframe.

Culprit-CZ:
Do as you wish. Few users wished to valid HTML

senky:
MA

Culprit-CZ:
MA

PR #213, opened 2013-02-06
Modifying of CPF via profile does not limit text for it's length

senky:
You can input text of any lenght to input/textbox generated via quick edit of CPF in profile, and it gets accepted.

It is because there is no check when copying text to the field. `maxlength` attribute should be added, using `max` option, to prevent this.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4699

### DIFF
--- a/root/socialnet/js/jquery.editable.js
+++ b/root/socialnet/js/jquery.editable.js
@@ -167,7 +167,7 @@ jQuery(document).ready(function($) {
 				th.off(opts.eventActivate);
 				var idx = obj.editID;
 				var _deac = true;
-				var _input = $('<input type="text" name="editable-' + idx + '" value="' + obj.edit + '" class="' + opts.inputClass + '" />').css(opts.cssInput);
+				var _input = $('<input type="text" name="editable-' + idx + '" value="' + obj.edit + '"' + ((obj.editMax) ? ' maxlength="' + obj.editMax + '"' : '') + ' class="' + opts.inputClass + '" />').css(opts.cssInput);
 
 				th.html(_input);
 				var _position = obj.origin.length;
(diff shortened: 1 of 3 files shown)


senky:
Well, not necessarly, because it does not get accepted when submitted, but yes, I will add it there.

woipi90:
ah ok just tried it.. it works fine..

root/socialnet/js/jquery.editable.js
```diff
@@ -167,7 +167,7 @@ jQuery(document).ready(function($) {
 				th.off(opts.eventActivate);
 				var idx = obj.editID;
 				var _deac = true;
-				var _input = $('<input type="text" name="editable-' + idx + '" value="' + obj.edit + '" class="' + opts.inputClass + '" />').css(opts.cssInput);
+				var _input = $('<input type="text" name="editable-' + idx + '" value="' + obj.edit + '"' + ((obj.editMax) ? ' maxlength="' : '') + obj.editMax + '"' + ' class="' + opts.inputClass + '" />').css(opts.cssInput);
```
Culprit-CZ: Should be there
`' + ((obj.editMax) ? ' maxlength="' + obj.editMax + '"' : '') + '`
otherwise will be placed some wrong code into HTML

root/socialnet/js/jquery.editable.js
```diff
@@ -211,7 +211,7 @@ jQuery(document).ready(function($) {
 						if (_t_height < 100) {
 							_t_height = 100;
 						}
-						_input = $('<textarea name="editable-' + idx + '" class="' + opts.inputClass + '" style="min-height:22px;max-height:100px"></textarea>').css(opts.cssText);
+						_input = $('<textarea name="editable-' + idx + '"' + ((obj.editMax) ? ' maxlength="' : '') + obj.editMax + '"' + ' class="' + opts.inputClass + '" style="min-height:22px;max-height:100px"></textarea>').css(opts.cssText);
```
Culprit-CZ: Should be there
`' + ((obj.editMax) ? ' maxlength="' + obj.editMax + '"' : '') + '`
otherwise will be placed some wrong code into HTML

woipi90:
now it seems to be fine..

Culprit-CZ:
MA

Kamahl19:
MA

PR #214, opened 2013-02-06
If there is no sutable result for user search, user is redirected to activitypage

senky:
In the script, there is a rule, which says, that if no result matches the search query, user is redirected to activitypage.php?search=<your query>. If no result is found, user should be noticed about that fact, or redirected to classic phpBB's search with "author" input prefilled and submitted.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4721

### DIFF
--- a/root/socialnet/activitypage.php
+++ b/root/socialnet/activitypage.php
@@ -98,6 +98,12 @@ if (!class_exists('socialnet_activitypage'))
 					case 'search':
 
 						$username = request_var('username', '', true);
+
+						if ( $username == '' || $username == $user->lang['SN_AP_SEARCH'] )
+						{
+							redirect( append_sid("{$phpbb_root_path}memberlist.$phpEx") );
+						}
+
 						$username_clean = utf8_clean_string($username);
 
 						$db_username = $db->sql_escape($username);
@@ -133,7 +139,7 @@ if (!class_exists('socialnet_activitypage'))
 						}
 						else
 						{
-							$redirect = append_sid("{$phpbb_root_path}activitypage.{$phpEx}", "search={$username}");
+							$redirect = append_sid("{$phpbb_root_path}search.{$phpEx}", 'author=' . urlencode($username) . ( (utf8_strlen($username) > 2) ? '*' : '' ) );
 						}
 						redirect($redirect);
 


woipi90:
so far so good..

but try to submit an empty searchfield.. i think you would be redirected to the search form with search... as a default value..

maybe you can redirect it but add something like:

if ($username == $user->lang['SN_AP_SEARCH'])
{
$username = "";
}

root/socialnet/activitypage.php
```diff
@@ -133,7 +133,7 @@ function sn_ap_cmp_username($expr)
 						}
 						else
 						{
-							$redirect = append_sid("{$phpbb_root_path}activitypage.{$phpEx}", "search={$username}");
+							$redirect = append_sid("{$phpbb_root_path}search.{$phpEx}", 'author=' . $username . ( (strlen($username) > 2) ? '*' : '' ) );
```
gooof: strlen = utf8_strlen(htmlspecialchars_decode($username))
long but required

senky: Hmm, username probably does not need any further filtering, as it comes from `request_var`. Or `url_encode` should be enough...

woipi90: ye but dont you wanna have it decoded for strlen..

) ); -> ));

gooof: oh yepp, decode only for db insert checks. small test. xD
but utf8_strlen is required if the username start with `Ä`
strlen = 2 or more
utf8_strlen = 1

senky:
> but try to submit an empty searchfield.. i think you would be redirected to the search form with search... as a default value..

If you focus the input, placeholder disappears. So it should not be possible to submit form with placeholder in the place. Whatsmore, "Search…" is valid username, so we cannot limit it.

woipi90:
only wrote this because on dev page it works like this.. try to submit an empty inputfield..

i think the best way would be that it wouldnt react..

senky:
> only wrote this because on dev page it works like this

No, it removes placeholder when you click on the input. It also does not get submitted, when the input is blank.
Does not it work for you like that? What is your browser?

woipi90:
no.. ff..

senky:
I am on FF 18 and placeholder gets removed, when I click on the input. It remains when I focus on it using tab, but it is at least selected, so that I can remove it easily. It is strange behaviour that your FF does not act as mine acts. Try to clear your browser's cache.

woipi90:
ah ok.. sry my fault i didnt tell u clearly.. i didnt submit the form via pressing enter but per pressing the search button..

so at first the placeholder is cleared
when i press the button the defaultvalue of the button is entered again and the form gets submitted with 'search...' as a value..

senky:
Ah, so we should add an event listener to submit event, removing placeholder when submitting clicking on the button. Thanks for your explanation.

woipi90:
anotherthing but because u are allready working on this.. i can remember a problem with submitting searchstrings containing "..
adding:

$username = str_replace('"', '', $username);

of course directly after request_var above ur changes..

would solve that.. am i right?

senky:
What about replacing JS placeholder with the native one? I know there will be some users to which placeholder will not be show due to old browser not supporting this attribute, but to write "Search..." into input next to search button is the same as not write it at all, IMHO. So maybe we could replace it with `placeholder="Search..."` attribute? @Culprit-CZ, @Kamahl19?

woipi90:
but then it would be nice if u can introduce it for status and comment input-fields too..
i remember that there has been a discussion about introducing it before.. but it hadnt been introduced due it is a html5 compliant..
but a workaround for using it exists..

senky:
The problem is, that non-native placeholders will never be perfect, because inputs usually can be submitted with the placeholder's string as valid text. Removing it from comments or statuses would cause less activity, as it asks users to react. But search option is function that is asked by users, so there is no problem using native placeholder there.

Kamahl19:
I think we should use jquery placeholder because of backward compatibility.

senky:
I described the problem above. Using jQuery placeholder results in inability to cover all possible search inputs. At least the one which is placeholder itself...

woipi90:
but i dont think anybody would name himselve search... or?

what about writing some longer desc.. like 'Searching for members' because then u wont have the problem eighter that phpbb users try to search the forum there..

fb for example uses something like 'Searching for members, places and things'..

senky:
True, but it is not wide enough to write such long string. But nevertheless, this is not part of this PR. So I will remake the script that it ignores "Search..." string.

woipi90:
maybe u can also add $username = str_replace('"', '', $username); to avoid the problems with the errornotices..

senky:
> maybe u can also add $username = str_replace('"', '', $username); to avoid the problems with the errornotices..

Error notices? What error notices? Whatever I do, whenever I put the `"` into the searched string, it simply gets encoded and parsed correctly.

Kamahl19:
MA

woipi90:
ok seems to be fine now..

Culprit-CZ:
MA

PR #215, opened 2013-02-06
IM chatbox does not stop blinking when focused by another than clicking method

senky:
If you have new message, chatbox should stop bliking when you focus on chatbox texarea. This, however, does not work for cases when you focus textarea by another way, than clicking. Also, when focus is already set to a chatbox, it should not start blinking at all.

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4738

### DIFF
--- a/root/socialnet/js/m.im.js
+++ b/root/socialnet/js/m.im.js
@@ -134,7 +134,7 @@
 				useEnter: false
 			});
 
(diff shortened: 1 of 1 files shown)


Culprit-CZ:
MA

Kamahl19:
MA

PR #216, opened 2013-02-07
WIP: Reduce SN load in notify

senky:
http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4742

### DIFF
--- a/root/socialnet/notify.php
+++ b/root/socialnet/notify.php
@@ -60,23 +60,30 @@ if (!class_exists('socialnet_notify'))
 		function init()
 		{
 			global $user, $db, $phpbb_root_path, $phpEx, $template, $config;
+
 			$this->time = time();
 
-			$this->ntf_delete();
+			if ($this->p_master->script_name == 'activitypage' && request_var('mode', false) == 'notify')
+			{
+				$this->ntf_delete();
 
-			$this->ntf_mark(SN_NTF_STATUS_UNREAD);
+				$this->ntf_mark(SN_NTF_STATUS_UNREAD);
 
-			$this->ntf_check_MARK();
+				$this->ntf_check_MARK();
 
-			$this->ntf_ap_show();
+				$this->ntf_ap_show();
+			}
 
-			$template->assign_vars(array(
-				'U_VIEW_NOTIFY'				 			=> append_sid("{$phpbb_root_path}activitypage.$phpEx", 'mode=notify'),
-				'S_SN_USER_UNREAD_NOTIFY'		=> $this->ntf_notify_count(),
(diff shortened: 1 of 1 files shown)


root/socialnet/notify.php
```diff
@@ -60,23 +60,30 @@ function socialnet_notify(&$p_master)
 		function init()
 		{
 			global $user, $db, $phpbb_root_path, $phpEx, $template, $config;
+
 			$this->time = time();
 
-			$this->ntf_delete();
+			if ($this->p_master->script_name == 'activitypage' && request_var('mode', false) == 'notify')
```
senky, line 66: Please, check for this condition. I am not sure, that it is enough to run those functions just in activitypage.php?mode=notify.

gooof, line 66: Its work on the `activitypage.php?mode=notify` page, but the `request_var('mode', false)` looks wrong, but its worked.
For 'notify' you need normally `request_var('mode', '')`

gooof, line 66: Now the link `profile.php?u=53&status_id=29&ntfMark=17#socialnet_us` is not longer working.
e.g. if you click on a link in the ajax notify box.

This is what ntf_check_MARK() doing. But is not better if simply call ntf_markID() here with `&ntfMark=17`?
ntf_markID need than only an user_id check and the notification was marked as read.

senky, line 66: @Culprit-CZ, please react on this comments. Is it possible to hide `ntf_check_MARK()` to the condition above, and call `ntf_markID()` independently when user accessed profile.php with `ntfMark` parametre included? (maybe call in the profile.php would solve the problem)

Culprit-CZ, line 66: Lines 174-175 should be
`$ntf_link[1] = 'ntfMark=' . $rowset[$i]['ntf_id'] . '&amp;' . $ntf_link[1];`
Is better and works

Culprit-CZ:
Can be merge problem with #202

senky:
I will rebase it if needed.

root/socialnet/notify.php
senky, line 66: It is not part of this PR. I just wonder if it is possible to create a condition that covers all possible situations. If it is, Culprit, you should be the only one who can create it, because I have no idea when and how are all those functions called. So please, either give it your time and post here how condition should look like, or I will close it, as we are not able to compose valid and working condition.

Kamahl19:
What is missing here?

gooof:
move line 72 `$this->ntf_check_MARK();` to line 77 before `if ( !isset($socialnet) || !defined('SN_NOTIFY') ) // not an ajax call`

PR #218, opened 2013-02-15
jQuery 1.9.1 + javascript minify

Culprit-CZ:
Due to problems encountered with the squash commits by #204 is necessary to complete the changes that have been made.

I apologize for my emerging issues.


woipi90:
how about including this one as well again.. https://github.com/podio/jquery-mentions-input/commit/5df262507feb00ac2e732203e188bf033e0ebbe0

root/styles/prosilver/template/socialnet/overall_header_hook_js.html
```diff
@@ -1,20 +1,20 @@
...
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.form.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.titlealert.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.form.min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.titlealert.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.expander.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.editable.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.jgrowl.min.js"></script>
-<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/underscore-min.js"></script>
+<script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/underscore.min.js"></script>
 <script type="text/javascript" src="{T_SOCIALNET_JS_PATH}/jquery.mentionsInput.js"></script>
```
senky: I believe this plugin can be minified as well.

Culprit-CZ: Sure, I'm going step by step.

senky:
Is the patch ready to be approved? Please, add "WIP" prependix if you are still working on the PR and remove it inly after your job is ready and you are waiting for our comments.

woipi90:
i think it includes all prior changes..

senky:
`jquery.elastic.source.js` is not minified. How much it differ from the original code? If it is only one or two changes, we can minify it and add a comment about modified files.

`jquery.events.input.js` are not minified, too. It differs only in one change. So please minify it and add a comment about the change (btw, the comment is added already there, so we can expect file to be up-to-date for jQuery UI 1.10.0 in next update.

woipi90:
btw ui 1.10.1 is released..

Culprit-CZ:
**`jquery.elastic.source.js`** is not original file because I put there 

``` js
var defaults = {
    showNewLine: true,
    useEnter : true,
    enterReplacement : '<br />',
    parentElement: null,
    submitElement: null
}
```

and for these options I have modified the plugin.
We can use minified file in releases. But we need to leave it not minified when we need to make some changes.

**`jquery.events.input.js`** is not original file also. There is need change for jQuery 1.9.0 [.addBack( selector ) replaces .andSelf()](http://jquery.com/upgrade-guide/1.9/#addback-selector-replaces-andself-).
I think it could be minified

senky:
So please, do it like this:
rename both files to have `.dev.js` appendix (`jquery.elastic.dev.js`, `jquery.events.input.dev.js`) and minify both of them for users.

Culprit-CZ:
Good Idea. 
What about put our dev files into specifics folder outside the main root folder

senky:
> What about put our dev files into specifics folder outside the main root folder

Great idea! And we will remove it when creating package for users. But please, firstly open a topic at the site, I have some commets for it.

senky:
MA

Kamahl19:
MA

PR #194, opened 2013-02-19
Unify deleting statuses and comments

senky:
When status is collapsed and you click Delete, it displays short status in confirm box.
When status is expanded, it displays the whole status in CB.

However, when comment is collapsed, it displays the whole comment in CB. I think this should be unified and only display collapsed comment. 

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4680

### DIFF
--- a/root/socialnet/js/m.socialnet.js
+++ b/root/socialnet/js/m.socialnet.js
@@ -381,7 +381,7 @@ var socialNetwork = (function($) {
 				});
 			}
 		},
-		
+
 		_minBrowser: function() {
 			var minBrowsers = {
 				msie: 8,
@@ -411,7 +411,7 @@ var socialNetwork = (function($) {
 
 			return true;
 		},
-		
+
 		_uaMatch: function(ua) {
 			ua = ua.toLowerCase();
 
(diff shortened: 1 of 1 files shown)


Culprit-CZ:
I think it is OK.
MA

Kamahl19:
MA

woipi90:
i just tried it on my profile on dev board and there it is still the same..
for statuses and comments..

senky:
I am not sure you understand, what does this script do. It should be the same. And...did you clear your browser cache?

woipi90:
i tested it on dev page and no in both cases it is handled different!

senky:
How different? please, post a screenshots, or describe it better, please.

woipi90:
first
second

senky:
What is wrong with this? I cannot see any problem...

woipi90:
is it supposed to act different when collapsed?
the two pictures show the difference on deleting.. its the same status i wanted to delete..
as u can see there isnt even shown the whole statustext when its collapsed..

senky:
Yes. If it is collapesed, only smaller part of a text should be displayed. If it is not collapsed, it should display full text in the prompt as well.

woipi90:
ok.. then it would be good to at least resize the expanded confirmbox..

PR #219, opened 2013-02-28
DEV javascript files

Culprit-CZ:
http://phpbbsocialnetwork.com/viewtopic.php?f=2&t=4895


dev-js/hooks/hook_dev-js.php
```diff
@@ -0,0 +1,40 @@
...
+	{
+		return;
+	}
+	
+	$JSpacker_path = $phpbb_root_path . '../dev-js/';
+
+	require_once $JSpacker_path . 'closureapi.php';
+
+	$closure = new closureCompiler();
+	$closure->setOptimalization('WHITESPACE_ONLY');
```
senky: SIMPLE_OPTIMIZATIONS can be used without worry.

dev-js/closure/closure_offline.php
```diff
@@ -0,0 +1,30 @@
...
+	public function compile()
+	{
+		global $phpbb_root_path;
+		
+		if ($this->script == '')
+		{
+			return;
+		}
+
+		$exec_cmd = 'java -jar ' . $phpbb_root_path . '../dev-js/closure/compiler.jar ' . $this->filename;
```
senky: I am not sure, but this does not look like generating the same output as .bat file does. Or this call at least does not look like naming files as `.min.js`.

dev-js/hooks/hook_dev-js.php
Culprit-CZ: Sure

dev-js/closure/closure_offline.php
Culprit-CZ: A have tested both.
How bat file so php script generate the same.

PR #200, opened 2013-03-07
Add "with " and "since " to relationships management

Kamahl19:
`<selection:married> with <inputfield> since <inputfield2>`

http://phpbbsocialnetwork.com/viewtopic.php?f=6&t=4736&p=20785#p20785


woipi90:
works for me..

Culprit-CZ:
MA

senky:
MA

PR #220, opened 2013-03-10
Adding friend from suggestions list does uncomplete job

senky:
- user is not redirected to suggestions, as written
- user is still displayed in the mutual friends list

http://phpbbsocialnetwork.com/viewtopic.php?f=44&t=4794


Kamahl19:
MA

Culprit-CZ:
MA

PR #221, opened 2013-03-10
purge_friends function lacks global variables

senky:
`purge_friends()` needs `$phpbb_root_path` and usage of `$this` instead of `$socialnet`.


woipi90:
should work then.. ;)

woipi90:
MA

